PR #1, opened 2026-03-05
Add gitcron integration: native tool + skills

patel-lyzr:
## Summary
- Add native `gitcron` built-in tool (`src/tools/gitcron.ts`) that wraps the gitcron CLI for scheduling, task management, and reminders
- Register gitcron tool in `src/tools/index.ts`
- Add three skills: `scheduling`, `task-management`, `reminders` with SKILL.md definitions
- Update `agent.yaml` to declare the gitcron tool and skills

## Test plan
- [ ] Verify `npm run build` compiles without errors
- [ ] Verify gitcron tool is registered and available in agent runtime
- [ ] Verify skills are loaded from `skills/` directory

Author: Khush Patel (khush@lyzr.ai) @patel-lyzr

### DIFF
--- a/agent.yaml
+++ b/agent.yaml
@@ -10,5 +10,10 @@ tools:
   - read
   - write
   - memory
+  - gitcron
+skills:
+  - scheduling
+  - task-management
+  - reminders
 runtime:
   max_turns: 50
--- a/skills/reminders/SKILL.md
+++ b/skills/reminders/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: reminders
+description: Set recurring or one-shot reminders that create GitHub issues on schedule — for compliance deadlines, reviews, follow-ups, and check-ins
+license: MIT
+allowed-tools: gitcron cli read write
+metadata:
+  author: open-gitagent
+  version: "1.0.0"
+  category: automation
+---
+
+# Reminders
+
+## Instructions
+
+You can create reminders that automatically create GitHub issues on a schedule or at a specific time. Reminders are defined in `cron.yaml` and compiled to GitHub Actions workflows.
+
+### Reminder types
+
+**Recurring** — Fires repeatedly on a cron schedule:
+```yaml
+- name: weekly-standup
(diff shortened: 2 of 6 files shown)


shreyas-lyzr (changes requested):
The gitcron integration adds scheduling skills and a native tool. However:

1. This modifies \`agent.yaml\` at the repo root — that file is a template/scaffold for new agents. Adding \`gitcron\` as a required tool there means every new agent gets it by default, which may not be desired.
2. The skills are well-structured but this should be opt-in via the plugin system or skills marketplace, not baked into the core scaffold.

Suggest making this a plugin or installable skill pack instead of modifying the core agent template.

shreyas-lyzr:
Apologies for the very late review @patel-lyzr.

State of play: `gitcron` is not in main today. The PR is technically still relevant, but a few questions before I'd merge:

1. **Is gitcron the canonical scheduling tool we want bundled as a built-in?** Today the codebase has `src/schedules.ts` and `src/schedule-runner.ts` doing scheduling internally. If those already cover the use cases this PR addresses, gitcron is redundant. If they cover *different* concerns (e.g., gitcron is for cross-agent scheduling, internal scheduling is for per-agent), then both can coexist — but the README / docs need to explain the split.
2. **Built-in vs optional?** Even if gitcron is the right tool, I'd lean toward shipping it as an installable skill/plugin rather than a built-in `src/tools/gitcron.ts`. The skill ecosystem (`skills/scheduling`, `skills/task-management`, `skills/reminders`) is exactly the right place for it — and you already created those skills in this PR. Worth considering whether the built-in tool wrapper adds enough on top.
3. **Rebase needed regardless** — two months of drift, 446 lines diff.

If gitcron is still the planned direction, please confirm and rebase; I'll merge same-day. If it makes more sense as a separate `gitclaw-gitcron-skill` plugin package, I'd close this PR and you could ship that independently.

Real apologies for the wait. Either direction is fine — just need a call on which.

PR #5, opened 2026-03-14
Implement steer() method for SDK query interface

pushkarsingh32:
## Summary
- Implement `steer()` on the `Query` interface using pi-agent-core's native `agent.steer()` API
- Hoists agent reference to outer scope so `steer()` can access it from the returned Query object
- Steer messages are queued and picked up between tool calls by the agent loop — no custom abort/re-prompt logic needed
- Add steer() unit tests and an example script for manual verification

## Changes
- `src/sdk.ts` — Hoist `_agent` ref, implement `steer()` with `UserMessage` format
- `test/sdk.test.ts` — Tests for steer (throws before agent loads, method exists on Query)
- `examples/test-steer.ts` — Manual SDK steer test that forces tool calls and steers after 3s

## Test plan
- [x] All 21 existing + new tests pass (`npx tsx --test test/sdk.test.ts`)
- [x] Manual test with `examples/test-steer.ts` confirms agent redirects mid-conversation
- [ ] Reviewer: verify steer works with multi-tool-call agents

### DIFF
--- a/examples/test-steer.ts
+++ b/examples/test-steer.ts
@@ -1,0 +1,80 @@
+/**
+ * Test the steer() SDK method.
+ *
+ * Strategy: Use a prompt that forces tool usage (e.g. "read a file"),
+ * so the agent loop stays alive long enough for steer() to be picked up
+ * between tool calls.
+ *
+ * Usage:
+ *   export OPENAI_API_KEY="sk-..."
+ *   node --experimental-strip-types --experimental-detect-module examples/test-steer.ts
+ */
+
+import { query } from "../dist/exports.js";
+
+async function main() {
+	const agentDir = new URL("../agents/assistant", import.meta.url).pathname;
+
+	console.log("Starting query...\n");
+	const q = query({
+		prompt: "Read the file SOUL.md and then read RULES.md and summarize both files.",
+		dir: agentDir,
+	});
+
+	let steered = false;
+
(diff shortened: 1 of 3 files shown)


shreyas-lyzr (changes requested):
The \`steer()\` SDK method is a nice addition — lets callers inject messages mid-conversation. However:

1. The PR adds \`tsx\` as a devDependency in package-lock.json — this pulls in esbuild with ~20 platform-specific optional deps. That's heavy for a dev tool. Consider using \`node --experimental-strip-types\` (already used in the test script) instead.
2. The example file imports from \`../dist/exports.js\` — should work without a build step if it's an example. Consider noting that \`npm run build\` is required first.

The core \`steer()\` implementation itself looks correct. Please drop the tsx dep and I'll approve.

shreyas-lyzr:
Apologies for letting this sit — and the apology is the real cost here, because the underlying design was right.

`steer()` is now in main (`src/sdk.ts` already has the `steer(_message: string)` method on the Query interface, backed by `pi-agent-core`'s `agent.steer()`). The implementation that landed is close to what you proposed — message queued, picked up between tool calls, no custom abort/re-prompt logic.

Closing this as superseded. Your spec materially shaped what shipped. Thank you for the patience and for the clean PR description — it made the implementation review easy when it finally happened.

PR #6, opened 2026-03-14
Rewrite CLI REPL with scroll region UI and live steering

pushkarsingh32:
## Summary
- Replace readline-based REPL with custom raw-mode terminal UI using ANSI scroll regions
- Input line is always visible at the bottom while streaming output scrolls above
- Users can type and send steer messages mid-stream without garbling output

## Layout
```
Rows 1..(r-3)  — scroll region (streaming output)
Row  r-2       — queue line (pending steer message in yellow)
Row  r-1       — separator
Row  r         — input line with fake block cursor
```

## Key changes
- Always-on raw mode with manual keystroke handling (no readline)
- Steer support: type during streaming, press Enter to send `agent.steer()`
- Queue line shows pending steer message, echoes to scroll region when agent picks it up (appears after AI response ends, not inline with streaming text)
- Hidden real cursor with inverse-video fake block cursor on input line
- Proper vertical spacing between user messages and AI responses
- Terminal resize handler redraws all fixed UI elements
- Header line tracking prevents scroll region from overwriting startup info

## Test plan
- [x] Build compiles clean
- [x] Manual test: typing during streaming shows on input line without garbling
- [x] Manual test: Enter during streaming sends steer, shown on queue line
- [x] Manual test: steer message appears in scroll region after AI's current response
- [x] Manual test: Ctrl+C aborts during streaming, exits when idle
- [x] Manual test: /quit, /memory, /skills commands work
- [ ] Reviewer: test on different terminal sizes and emulators

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -472,20 +472,21 @@ async function main(): Promise<void> {
 
 	agent.subscribe((event) => handleEvent(event, hooksConfig, agentDir, sessionId, auditLogger));
 
-	console.log(bold(`${manifest.name} v${manifest.version}`));
-	console.log(dim(`Model: ${loaded.model.provider}:${loaded.model.id}`));
+	let headerLines = 0;
+	console.log(bold(`${manifest.name} v${manifest.version}`)); headerLines++;
+	console.log(dim(`Model: ${loaded.model.provider}:${loaded.model.id}`)); headerLines++;
 	const allToolNames = tools.map((t) => t.name);
-	console.log(dim(`Tools: ${allToolNames.join(", ")}`));
+	console.log(dim(`Tools: ${allToolNames.join(", ")}`)); headerLines++;
 	if (skills.length > 0) {
-		console.log(dim(`Skills: ${skills.map((s) => s.name).join(", ")}`));
+		console.log(dim(`Skills: ${skills.map((s) => s.name).join(", ")}`)); headerLines++;
 	}
 	if (loaded.workflows.length > 0) {
-		console.log(dim(`Workflows: ${loaded.workflows.map((w) => w.name).join(", ")}`));
+		console.log(dim(`Workflows: ${loaded.workflows.map((w) => w.name).join(", ")}`)); headerLines++;
 	}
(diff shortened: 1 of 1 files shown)


shreyas-lyzr (changes requested):
This is a large rewrite of the CLI REPL — raw mode, scroll regions, custom input handling. Concerns:

1. **Very large diff** touching the core REPL loop. This replaces readline entirely with raw terminal management. High risk of regressions on different terminal emulators (iTerm, Windows Terminal, SSH sessions, etc.)
2. **No fallback** — if raw mode fails (e.g. piped input, CI, non-TTY), the entire CLI breaks. Need a \`process.stdin.isTTY\` guard with fallback to the current readline approach.
3. The \`headerLines\` counter is fragile — any log line change breaks the scroll region math.

I'd suggest splitting this: first PR for the steer queue (simple feature), second PR for the scroll region UI (risky, needs testing).

shreyas-lyzr:
Apologies for the long silence.

The underlying design — ANSI scroll regions, always-visible input row, queued steer messages mid-stream — still reads as the right call. The CLI today still uses `readline` (`src/index.ts: import { createInterface } from "readline"`), so the rewrite would replace real code, not phantom.

**Two things have changed since March that affect this:**

1. **`steer()` shipped on the SDK** (a different PR landed it). The CLI doesn't yet expose it interactively — your raw-mode UI with the queue line is exactly the missing piece. Worth keeping.
2. **Cloud-mode prompt + username login + in-UI HTML/PDF artifact previews** landed on May 5. These add new things the REPL needs to render — your rewrite needs to handle them or coexist cleanly.

**Two paths:**

1. **Rebase and confirm the UX still works** with cloud-mode prompts and artifact previews. If so, ship it.
2. **Decompose:** scroll-region + queue line in one PR, mid-stream input handling in a follow-up. Easier review, easier to rebase if main moves again.

Either way, this is a real UX upgrade — the current REPL drops keystrokes when streaming. Just need a fresh diff. Will review within a day of rebase.

Apologies again for letting this drift.

PR #15, opened 2026-03-17
feat: auto-detect API keys from env vars for zero-config startup

### LINKED ISSUE #14: Auto-detect API keys and default to latest models for zero-config startup
## Problem                                                                                                                                                                      
                                                                                                                                                                                  
  Currently, `gitclaw` requires explicit model configuration to start. If a user has API keys set in their shell (e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`) or in a `.env` file,
   gitclaw does not detect them or pick a default model automatically. Instead it fails with:                                                                                     
                                                                                                                                                                                  
  No model configured. Either:                                                                                                                                                    
  - Set model.preferred in agent.yaml (e.g., "anthropic:claude-sonnet-4-5-20250929")                                                                                              
  - Pass --model provider:model on the command line                                                                                                                               
  
  Users must always either edit `agent.yaml` or pass `--model provider:model` manually, even when valid API keys are already available in the environment.                        
                                                                                                                                                                                  
  Additionally, there is no `.env` file support — API keys must be exported in the shell.                                                                                         
                                                                                                                                                                                  
  ## Expected Behavior                                                                                                                                                            
                                                                                                                                                                                
  `gitclaw` should work out of the box when an API key is present:                                                                                                                
  
  1. **Auto-detect available API keys** — On startup, check for known environment variables (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GOOGLE_API_KEY`, `XAI_API_KEY`,              
  `GROQ_API_KEY`, `MISTRAL_API_KEY`)                                                                                                                                            
  2. **Pick a sensible default model** — If no model is configured in `agent.yaml` or via `--model`, automatically select the latest/best model for the first available API key.  
  Suggested defaults:                                                                                                                                                             
     - `ANTHROPIC_API_KEY` → `anthropic:claude-sonnet-4-5-20250929`
     - `OPENAI_API_KEY` → `openai:gpt-4o`                                                                                                                                         
     - `GOOGLE_API_KEY` → `google:gemini-2.0-flash`                                                                                                                               
     - (etc.)                                                                                                                                                                     
  3. **Support `.env` files** — Load `.env` from the agent directory (and/or home directory) so users can persist their API keys without exporting in every shell session         
                                                                                                                                                                                  
  The resolution priority should remain:                                                                                                                                          
  --model flag > agent.yaml model.preferred > env config override > auto-detected from API key                                                                                    
                                                                                                                                                                                  
  ## Current Code                                                                                                                                                               
                                                                                                                                                                                  
  **API key validation** (`src/index.ts:468-484`) already has the provider-to-env-var mapping:                                                                                    
  
  ```typescript                                                                                                                                                                   
  const apiKeyEnvVars: Record<string, string> = {                                                                                                                               
    anthropic: "ANTHROPIC_API_KEY",                                                                                                                                               
    openai: "OPENAI_API_KEY",
    google: "GOOGLE_API_KEY",                                                                                                                                                     
    xai: "XAI_API_KEY",                                                                                                                                                         
    groq: "GROQ_API_KEY",                                                                                                                                                         
    mistral: "MISTRAL_API_KEY",                                                                                                                                                 
  };                                                                                                                                                                              
 ```                                                                                                                                                                       
  But this is only used to validate after a model is already chosen — not to auto-detect which provider to use.                                                                   
                                                                                                                                                                                  
  Model resolution (src/loader.ts:354-360) throws an error when no model is configured instead of falling back to auto-detection.                                                 
                                                                                                                                                                                
  Suggested Fix                                                                                                                                                                   
                                                                                                                                                                                
  1. Add .env loading — Use dotenv (or Node's --env-file) to load .env from the agent directory early in startup                                                                  
  2. Add auto-detection in loadAgent() — When no model is configured via any existing method, iterate through apiKeyEnvVars and pick the first available provider with a sensible
  default model                                                                                                                                                                   
  3. Log the choice — Print which model was auto-selected so the user knows (e.g. Using anthropic:claude-sonnet-4-5-20250929 (detected ANTHROPIC_API_KEY))

Jah-yee:
## Summary

Adds auto-detection of API keys from environment variables (shell env or .env file) to enable zero-configuration startup. When no model is explicitly configured, gitclaw now checks for common API key environment variables and automatically selects an appropriate default model.

## Changes

- Add `autoDetectModelFromEnv()` function that checks for common API key env vars
- Supports: `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GOOGLE_API_KEY`, `OLLAMA_HOST`, `GROQ_API_KEY`, `MISTRAL_API_KEY`, `DEEPSEEK_API_KEY`, `XAI_API_KEY`, and more
- Also checks `.env` file in the current directory
- Maps each API key to a sensible default model (e.g., `ANTHROPIC_API_KEY` -> `claude-sonnet-4-5`)
- Updates error message to mention the auto-detection option

## Usage

Users can now run gitclaw without any configuration if they have an API key set:

```bash
export ANTHROPIC_API_KEY=sk-...
gitclaw # Automatically uses claude-sonnet-4-5
```

Or via .env file:

```bash
# .env
ANTHROPIC_API_KEY=sk-...
```

Fixes #14

### DIFF
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -1,4 +1,5 @@
 import { readFile, mkdir, writeFile } from "fs/promises";
+import { readFileSync } from "fs";
 import { join } from "path";
 import { randomUUID } from "crypto";
 import { execSync } from "child_process";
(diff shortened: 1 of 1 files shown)


shreyas-lyzr (changes requested):
Good idea — zero-config startup is a nice DX win. A few issues to address before merging:

## Bugs

1. **Typo: `ANYSACLE_API_KEY`** should be `ANYSCALE_API_KEY` — this will silently never match.

2. **`.env` reading uses `cwd()`, not the agent directory.** `readFileSync(".env")` reads from the process working directory, which may not be the agent dir. This should use the agent directory path (consistent with how `.env` is loaded elsewhere in the codebase). However, the agent dir isn't available at this call site — see design concern below.

3. **Duplicate iteration logic.** When a `.env` file exists, the `process.env` keys are checked inside the try block. If no `.env` exists, they're checked again after the catch. This means when `.env` *does* exist but doesn't contain any matching keys, `process.env` keys are still checked in the first loop — so the second loop is dead code in that path. Simpler to load `.env` vars into a merged object once, then iterate once.

4. **`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` map to the same model separately.** Having just an access key without a secret (or vice versa) would falsely trigger auto-detection. These should be checked as a pair (`&&` not `||`).

## Design concerns

5. **Stale model IDs.** Hardcoded model IDs like `claude-sonnet-4-5-20250505` and `gpt-4o-2024-11-20` will go stale. Consider using versionless aliases (`claude-sonnet-4-5`, `gpt-4o`) or pulling defaults from a config, so this doesn't need code changes every model release.

6. **`.env` parsing doesn't handle quotes or comments.** Lines like `ANTHROPIC_API_KEY="sk-..."` will include the quotes in the value. Lines starting with `#` will be parsed as key-value pairs. The existing `.env` parser in `server.ts` has the same minimal parsing but at least trims — consider reusing that logic or adding quote stripping and comment handling.

7. **Function placement.** This auto-detection runs in `loadAgent()` which doesn't receive the agent directory. The `.env` file should be read relative to the agent dir, not cwd. You may want to accept the agent dir as a parameter, or move the `.env` loading to the caller.

Jah-yee:
I’m closing this to clean up my open queue; there was no follow-up. If a maintainer wants the idea, I can reopen or resubmit. Thanks for the project either way.

Issue #48, opened 2026-06-01
Auto Model Routing for Autonomous Agents to Reduce Token Usage and Cost


harshit-lyzr:
<html><head></head><body><h3>Description</h3><h4>Problem</h4><p>Autonomous agents currently use a single model for their entire execution flow. When a reasoning-capable model is selected, it is applied to every task, including simple operations such as summarization, extraction, formatting, and transformations.</p><p>This results in:</p><ul><li><p>Unnecessary token consumption</p></li><li><p>Higher inference costs</p></li><li><p>Increased latency for lightweight operations</p></li><li><p>Inefficient utilization of model capabilities</p></li></ul><p>Many agent workflows contain a mix of simple and complex tasks, but today all tasks are executed using the same model regardless of complexity.</p><hr><h4>Proposed Solution</h4><p>Introduce <strong>Automatic Model Routing</strong> within autonomous agents.</p><p>The agent runtime should classify tasks based on complexity and automatically route them to the most appropriate model:</p>
Task Type | Recommended Model
-- | --
Summarization | Lightweight model
Data extraction | Lightweight model
Classification | Lightweight model
Text transformation | Lightweight model
File search & analysis | Reasoning model
Multi-step planning | Reasoning model
Decision-making | Reasoning model
Tool orchestration | Reasoning model
Complex problem solving | Reasoning model

[...]

PR #51, opened 2026-06-01
feat: auto model routing for SkillFlows

RiteshTiwari1:
Addresses #48

## Summary

Adds **automatic model routing** for SkillFlows. Each step is classified by
complexity at runtime and routed to the right model — lightweight tasks
(summarize / extract / classify / transform) run on a cheaper model, while
reasoning-intensive tasks (planning / decision-making / tool orchestration)
stay on the configured reasoning model. Reduces token usage and cost without
sacrificing quality on complex steps.

## How it works

Model resolution per step, in priority order:

1. Explicit per-step `model:` in `workflows/*.yaml`
2. Per-skill `model:` in `SKILL.md`
3. Automatic classification via `model.routing` tiers
4. Fallback to the `preferred` model

Unknown / ambiguous tasks default to **reasoning**, so cost optimization never
silently degrades quality. Routing is **opt-in** — active only when a
`model.routing` block is present, so existing agents are unaffected.

## Configuration (`agent.yaml`)

```yaml
model:
  preferred: "openai:gpt-5-reasoning"
  routing:
    lightweight: "openai:gpt-4o-mini"
    reasoning:   "openai:gpt-5-reasoning"
```

## Scope note

Routing is applied at the **SkillFlow step** level — where GitAgent runs
discrete tasks in sequence. Per-turn routing inside a single agent loop would
need deeper `pi-agent-core` changes; happy to follow up if that's preferred.

### DIFF
--- a/src/model-routing.ts
+++ b/src/model-routing.ts
@@ -1,0 +1,143 @@
+// Auto Model Routing (issue #48)
+//
+// Classifies each task in an agent workflow by complexity and routes it to the
+// most appropriate model: lightweight tasks (summarize/extract/classify/
+// transform) go to a cheap model, while reasoning-intensive tasks (search,
+// planning, decision-making, tool orchestration, complex problem solving)
+// stay on the configured reasoning model. Explicit per-step / per-skill model
+// settings always win, and anything unresolved falls back to the primary model.
+
+export type ModelTier = "lightweight" | "reasoning";
+
+export interface RoutingConfig {
+	/** Master switch. Defaults to true when a routing block is present. */
+	enabled?: boolean;
+	/** Concrete model id for lightweight tasks, e.g. "openai:gpt-4o-mini". */
+	lightweight?: string;
+	/** Concrete model id for reasoning tasks, e.g. "openai:gpt-4o". */
+	reasoning?: string;
+	/** User overrides for classification — first matching rule wins. */
+	rules?: Array<{ tier: ModelTier; match: string[] }>;
+}
+
+export interface RouteInput {
+	/** Explicit per-step model (highest priority). May be an alias or model id. */
+	stepModel?: string;
+	/** Per-skill default model from SKILL.md frontmatter. May be an alias or id. */
+	skillModel?: string;
+	/** Text used to classify the task (typically skill name + step prompt). */
+	classifyText: string;
+	/** Routing configuration from agent.yaml (model.routing). */
+	routing?: RoutingConfig;
+	/** The agent's primary/preferred model — the ultimate fallback. */
+	primaryModel?: string;
+}
+
+export interface RouteResult {
+	/** Resolved concrete "provider:model" string (undefined → let runtime decide). */
+	model?: string;
+	/** The complexity tier, when the model came from automatic classification. */
+	tier: ModelTier | null;
+	/** Where the decision came from. */
+	source: "step" | "skill" | "auto" | "fallback";
+}
+
+// Default task-to-tier keyword framework, derived directly from the issue's
+// recommended task-type table. Matched against word starts so "summarize",
+// "summary" and "summarization" all hit "summ", without false positives like
+// "already" matching "read".
+const DEFAULT_LIGHTWEIGHT = [
+	"summ", "extract", "classif", "transform", "format", "convert",
+	"parse", "fetch", "read", "load", "lookup", "normaliz", "translat",
+	"rephrase", "rewrite", "tag", "label", "render",
+];
+const DEFAULT_REASONING = [
+	"search", "analy", "plan", "decid", "decision", "orchestrat", "solve",
+	"reason", "validat", "evaluat", "review", "audit", "diagnos", "debug",
+	"architect", "design", "strateg", "investigat", "assess", "judge",
+	"verify", "critique", "infer", "deduc",
+];
+
+function matchesAny(text: string, keywords: string[]): boolean {
+	for (const kw of keywords) {
+		// Word-start boundary: keyword must begin a word.
+		const re = new RegExp(`\\b${kw.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}`, "i");
+		if (re.test(text)) return true;
+	}
+	return false;
+}
+
+/**
+ * Classify a task into a complexity tier. User-defined rules (from
+ * model.routing.rules) take precedence over the built-in defaults. When a task
+ * matches neither — or matches both — it defaults to "reasoning" so that
+ * reasoning quality is never sacrificed to save cost.
+ */
+export function classifyTaskTier(
+	classifyText: string,
+	rules?: Array<{ tier: ModelTier; match: string[] }>,
+): ModelTier {
+	const text = classifyText || "";
+
+	// User overrides first, in declaration order.
+	if (rules) {
+		for (const rule of rules) {
+			if (Array.isArray(rule.match) && matchesAny(text, rule.match)) {
+				return rule.tier;
+			}
+		}
+	}
+
+	const hasReasoning = matchesAny(text, DEFAULT_REASONING);
+	if (hasReasoning) return "reasoning";
+	const hasLightweight = matchesAny(text, DEFAULT_LIGHTWEIGHT);
+	if (hasLightweight) return "lightweight";
+
+	// Unknown → keep quality high.
+	return "reasoning";
+}
+
+/**
+ * Resolve a model reference that may be a routing-tier alias
+ * ("lightweight"/"reasoning") or a literal "provider:model" id.
+ */
+export function resolveModelAlias(ref: string | undefined, routing?: RoutingConfig): string | undefined {
+	if (!ref) return undefined;
+	if (ref === "lightweight") return routing?.lightweight || undefined;
+	if (ref === "reasoning") return routing?.reasoning || undefined;
+	return ref;
+}
+
+/**
+ * Decide which model a task should run on. Precedence:
+ *   1. explicit per-step model      (source: "step")
+ *   2. per-skill declared model      (source: "skill")
+ *   3. automatic classification      (source: "auto")  — when routing is enabled
+ *   4. primary/preferred model       (source: "fallback")
+ *
+ * Automatic routing is active only when a routing block is present and not
+ * disabled. If classification picks a tier with no configured model, it falls
+ * through to the primary model (fallback on routing failure).
+ */
+export function resolveRoutedModel(input: RouteInput): RouteResult {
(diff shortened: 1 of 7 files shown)


Nivesh353:
Hi @RiteshTiwari1 — apologies for the lack of review, and thanks for putting this together. The routing logic in model-routing.ts looks solid — clean priority chain, and defaults to the reasoning tier when a task is ambiguous so cost savings don't quietly hurt quality.

One thing that's changed since this was opened: src/voice/server.ts, where the routing gets wired into the SkillFlow execution loop, was removed in the 2.0 split (#56) — voice moved into its own @open-gitagent/voice package. That's why this now shows conflicts. The loader.ts, skills.ts, and workflows.ts changes still apply cleanly; it's just the execution wiring that needs a new home.

Let us know if you'd like to pick this back up against the new structure — happy to help figure out where it should hook in. Thanks again for the contribution.

RiteshTiwari1:
Thanks @Nivesh353. Rebased onto main (v2.0.2) and dropped the `src/voice/server.ts` changes, so the conflict's cleared. This PR now has: `model-routing.ts` (engine), the loader/skills/workflows config changes, `exports.ts` (re-exports `resolveRoutedModel` for the voice package), and unit tests.

The per-step wiring belongs in `executeFlow()` in `@open-gitagent/voice` — swap `query({ model: opts.model })` for `resolveRoutedModel(...)`. That's a follow-up importing from this change.

Two questions:

1. Should I open the voice-side PR, or will the team take it?
2. Keep `model-routing.ts` in core, or move it into the voice package (its only consumer)?

Nivesh353:
Thanks @RiteshTiwari1 

1. Please go ahead and open the voice-side PR yourself — you clearly know
   this code well, and it's a small, well-scoped swap per what you
   described. Happy to review it as soon as it's up.

2. Keep model-routing.ts in core. Voice isn't going to be its only
   consumer for much longer.

src/model-routing.ts
```diff
@@ -0,0 +1,115 @@
...
+	tier: ModelTier | null;
+	source: "step" | "skill" | "auto" | "fallback";
+}
+
+// Matched against word starts, so "summarize"/"summary"/"summarization" all hit
+// "summ" without "already" matching "read".
+const DEFAULT_LIGHTWEIGHT = [
+	"summ", "extract", "classif", "transform", "format", "convert",
+	"parse", "fetch", "read", "load", "lookup", "normaliz", "translat",
+	"rephrase", "rewrite", "tag", "label", "render",
```
shreyas-lyzr, line 64: Blocking: the keyword-prefix classifier is fundamentally fragile for this use case.

The core problem is that task complexity is a semantic property that does not reliably map to a fixed prefix vocabulary. Real SkillFlow prompts will break this in both directions:

Lightweight classified as reasoning (false-positive tax):
- "verify the file exists" → matches 'verif' → reasoning (but this is a trivial existence check)
- "review the spelling in this string" → matches 'review' → reasoning
- "validate that the field is not empty" → matches 'validat' → reasoning
- "search for the word hello in the file" → matches 'search' → reasoning

Reasoning classified as lightweight (silent quality downgrade, the worse case):
- "read the code and identify bugs" → matches 'read' → lightweight (but this is a debugging task)
- "fetch user data and detect anomalies" → matches 'fetch' → lightweight (complex analysis)
- "parse and understand the business requirements" → matches 'parse' → lightweight
- "load the config, then decide which deployment strategy to use" → matches 'load' → lightweight

The `read` and `fetch` entries in DEFAULT_LIGHTWEIGHT are especially risky: they fire on the first verb in a compound prompt, classifying the whole task by only part of it.

Suggested fix — replace classifyTaskTier with a single cheap model call:

```typescript
async function classifyTaskTierLLM(
  classifyText: string,
  lightweightModel: string,   // the cheap model is appropriate to classify itself
): Promise<ModelTier> {
  const prompt = [
    "Classify the following task as either 'lightweight' (summarize, extract, format,",
    "convert, parse, fetch, render — mechanical, no reasoning required) or 'reasoning'",
    "(plan, analyze, decide, debug, review, validate logic, orchestrate — needs judgment).",
    "Reply with exactly one word: lightweight or reasoning.",
    "",
    "Task: " + classifyText,
  ].join("\n");

  const response = await callModel(lightweightModel, prompt, { max_tokens: 5 });
  const word = response.trim().toLowerCase();
  return word === "lightweight" ? "lightweight" : "reasoning"; // unknown → reasoning (safe default)
}
```

This costs one gpt-4o-mini call per auto-classified step (~0.0001 USD at current pricing), which is negligible compared to the step itself, and it handles compound prompts, novel verbs, and context correctly. The existing user-rule override path (`rules:`) can stay as-is for deterministic overrides where operators want guaranteed behavior.

If a synchronous API is required (e.g. the call site can't be made async), the keyword approach is acceptable as a degraded fallback, but the DEFAULT_LIGHTWEIGHT list should at minimum remove 'read', 'fetch', 'load', and 'search' — these verbs are too context-sensitive to use as tier signals.

src/model-routing.ts
```diff
@@ -0,0 +1,115 @@
...
+];
+const DEFAULT_REASONING = [
+	"search", "analy", "plan", "decid", "decision", "orchestrat", "solve",
+	"reason", "validat", "evaluat", "review", "audit", "diagnos", "debug",
```
shreyas-lyzr, line 68: The 'search' keyword in DEFAULT_REASONING catches all uses of the word — including trivial lookup/grep steps that should route to the lightweight model. Consider removing it or narrowing it to more specific terms like 'research', 'investigate', or 'deep-search'. As-is, any step prompt containing 'search' (e.g. 'search the file for pattern X', 'full-text search the index') pays the reasoning model rate.

src/model-routing.ts
```diff
@@ -0,0 +1,115 @@
...
+	if (matchesAny(text, DEFAULT_LIGHTWEIGHT)) return "lightweight";
+	return "reasoning";
+}
+
+/** Resolve a tier alias ("lightweight"/"reasoning") or pass a model id through. */
+export function resolveModelAlias(ref: string | undefined, routing?: RoutingConfig): string | undefined {
+	if (!ref) return undefined;
+	if (ref === "lightweight") return routing?.lightweight || undefined;
+	if (ref === "reasoning") return routing?.reasoning || undefined;
+	return ref;
```
shreyas-lyzr, line 155: Silent fallback when a tier alias has no backing model.

If a SKILL.md specifies `model: lightweight` but the agent.yaml routing block omits `lightweight:`, `resolveModelAlias` returns `undefined`, and the caller falls through to `primaryModel` without any log or warning. The operator's explicit intention ("use a lightweight model for this skill") is silently ignored.

Consider emitting a warning — or throwing — when an alias resolves to undefined:

```typescript
export function resolveModelAlias(ref: string | undefined, routing?: RoutingConfig): string | undefined {
  if (!ref) return undefined;
  if (ref === "lightweight") {
    if (!routing?.lightweight) {
      console.warn(`[routing] tier alias 'lightweight' used but routing.lightweight is not configured; falling through`);
    }
    return routing?.lightweight || undefined;
  }
  if (ref === "reasoning") {
    if (!routing?.reasoning) {
      console.warn(`[routing] tier alias 'reasoning' used but routing.reasoning is not configured; falling through`);
    }
    return routing?.reasoning || undefined;
  }
  return ref;
}
```

test/model-routing.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+test("disabled routing falls back to primary", () => {
+	const r = resolveRoutedModel({
+		classifyText: "summarize the pull request",
+		routing: { ...routing, enabled: false },
+		primaryModel: "openai:gpt-5-reasoning",
+	});
+	assert.equal(r.model, "openai:gpt-5-reasoning");
+	assert.equal(r.source, "fallback");
+});
```
shreyas-lyzr, line 158: The test suite only covers the pure functions in model-routing.ts. There is no test that wires a SkillFlowStep with a model field through loadFlowDefinition/discoverWorkflows and verifies that the model field is preserved and passed to resolveRoutedModel correctly. A roundtrip test (serialize → deserialize via saveFlowDefinition → loadFlowDefinition, assert step.model survives) would catch regressions in the YAML plumbing.

shreyas-lyzr (changes requested):
Four issues across the new model-routing module. The most significant is the classifier design — covered in the inline comments below.

1. Keyword classifier produces systematic misclassifications (blocking — see inline).
2. 'search' is bucketed as reasoning; common lightweight use (grep/find/lookup) will always overpay.
3. resolveModelAlias returns undefined when a tier alias is configured but the routing block omits that tier's model string — silently skips the tier instead of warning. This can silently downgrade to fallback with no observability.
4. No integration/end-to-end test exercises resolveRoutedModel being called from actual SkillFlow execution — the unit tests cover the logic in isolation but there's no test proving the plumbing from workflows.ts through to the routed model call.

Security pass: no CVEs in changed deps (no new packages added), no hardcoded secrets, no injection surface — the classifyText input is only used in a regex match, not in any eval/exec/query path.

RiteshTiwari1:
Thanks @shreyas-lyzr , all four are fair.

2, 3, 4 I'll fix: drop search from the reasoning defaults, warn when a tier alias resolves to no configured model, and add a save→load roundtrip test for step model.

On 1: agreed keywords can't capture semantic complexity. Quick check before I rework —> an LLM classify call has to be async, so it'd live in voice's `executeFlow()` rather than the pure resolver in core (adds a per-step call on the cheap model). Good with that split, or would you rather the classifier stay in core behind an injected query fn?

Nivesh353:
Keep it in core behind an injected query fn, not voice  @RiteshTiwari1

RiteshTiwari1:
@shreyas-lyzr @Nivesh353 pushed a commit addressing all four.

* Classifier is now a single LLM call on the lightweight model, with the keyword list kept only as an offline fallback. Left it in core behind an injected query fn per @Nivesh353, and constrained it to one turn with no tools so it stays one cheap call.
* Dropped bare search from the reasoning defaults — research/investigate still route to reasoning.
* resolveModelAlias now warns when a tier alias has no configured model instead of silently falling through.
* Added a save → load roundtrip test that asserts step model survives, plus classifier tests with a fake injected query.

One heads-up: resolveRoutedModel is async now, so the voice follow-up will await it and pass query.

Nivesh353:
@RiteshTiwari1  We're currently in the phase of moving executeFlow() into core — will look into this once that PR is merged, since there may be some changes you'll need to make on your end depending on how it lands.

Tracking here if you want to follow along: https://github.com/open-gitagent/gitagent/pull/90

PR #59, opened 2026-06-18
Feat/mcp client support

Nivesh353:
## What
Adds **MCP (Model Context Protocol) client** support. Declare servers in `agent.yaml` (or via the SDK `mcpServers` option) and the agent automatically gains those servers' tools — no integration code.

## Why
Previously, every external capability (GitHub, Postgres, Slack, filesystem…) had to be hand-built and maintained by us as a custom tool/plugin. With MCP, users plug into the whole ecosystem of ready-made servers via a few lines of config, and we maintain none of it. Keeps gitagent aligned with the industry standard (Claude, Cursor, etc.).

## How it works
On agent load, gitagent connects to each declared server (stdio / HTTP / SSE), calls `listTools`, and registers every tool as a native `AgentTool` named `<server>__<tool>`. Execution forwards the call to the server and returns its result. Connections are pooled for the session and torn down on every exit path.


test/mcp.test.ts
```diff
@@ -0,0 +1,167 @@
...
+	});
+	it("prefers text content over structuredContent when both exist", () => {
+		assert.equal(
+			flattenToolResult({ content: [{ type: "text", text: "hi" }], structuredContent: { x: 1 } }),
+			"hi",
+		);
+	});
+});
+
+describe("buildToolsForConnection", () => {
```
shreyas-lyzr, line 83: The execute() return value is a plain string (flattenToolResult's output), not an MCP content array. These assertions will fail at runtime:

```ts
assert.equal(res.content[0].text, "hello");
```

Should be:

```suggestion
const res = await echo.execute("call-1", { msg: "hello" });
assert.equal(res, "hello");
```

Same issue on the boom and pic assertions below — they check .content[0].text against a string.

shreyas-lyzr (commented):
Solid MCP client integration — the architecture is well-thought-out: lazy loading, fail-soft per-server, idempotent cleanup on every exit path, and a proper test suite exercising the in-memory transport. A few issues worth addressing before merge.

Two correctness issues: the test assertions are verifying the wrong shape (checking `.content[0].text` on a value that `execute()` returns as a plain string, not an MCP result object), and the `withTimeout` timer is not cancelled on success, which accumulates unresolved timer handles in long-running sessions. One security concern: env var interpolation uses `|| ""` which silently replaces unset vars with empty string — a misconfigured auth token header becomes an empty string with no warning. Detailed notes inline.

src/mcp/manager.ts
```diff
@@ -0,0 +1,254 @@
...
+	name: string;
+	client: Client;
+	close: () => Promise<void>;
+}
+
+function withTimeout<T>(op: Promise<T>, ms: number, label: string): Promise<T> {
+	return Promise.race([
+		op,
+		new Promise<T>((_, reject) =>
+			setTimeout(() => reject(new Error(`${label} timed out after ${ms}ms`)), ms),
```
shreyas-lyzr: The timeout timer is never cleared when the promise resolves first. In a long-running REPL session with many tool calls this accumulates unresolved handles and can prevent clean process exit.

```suggestion
function withTimeout<T>(op: Promise<T>, ms: number, label: string): Promise<T> {
	return new Promise<T>((resolve, reject) => {
		const timer = setTimeout(
			() => reject(new Error(`${label} timed out after ${ms}ms`)),
			ms,
		);
		op.then(
			(v) => { clearTimeout(timer); resolve(v); },
			(e) => { clearTimeout(timer); reject(e); },
		);
	});
}
```

src/env-utils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Replaces `${VAR_NAME}` occurrences in strings with `process.env.VAR_NAME`,
+ * or an empty string when the variable is unset (matching the long-standing
+ * plugin-config behavior). Recurses through arrays and plain objects so an
+ * entire config object (e.g. an MCP server definition) can be interpolated in
+ * one call. Non-string leaf values are returned unchanged.
+ */
+const ENV_VAR_PATTERN = /\$\{(\w+)\}/g;
+
+export function interpolateEnvString(value: string): string {
+	return value.replace(ENV_VAR_PATTERN, (_, envName) => process.env[envName] || "");
```
shreyas-lyzr: Using `|| ""` silently drops unset variables. For most config values that is fine, but for auth headers (e.g. `Authorization: Bearer ${ANALYTICS_TOKEN}`) a missing env var becomes `Authorization: Bearer ` — the request goes out with a malformed header and no warning logged.

Consider warning when the var is unset and the caller is likely a secret:

```suggestion
export function interpolateEnvString(value: string): string {
	return value.replace(ENV_VAR_PATTERN, (_, envName) => {
		if (process.env[envName] === undefined) {
			console.warn(`[mcp] env var \${envName} is not set; substituting empty string`);
		}
		return process.env[envName] ?? "";
	});
}
```

Alternatively, keep it silent but document the behavior clearly in the README so users know a misconfigured token gives an empty string rather than an error.

src/mcp/manager.ts
```diff
@@ -0,0 +1,254 @@
...
+				break;
+			}
+			case "resource_link":
+				parts.push(`[resource_link: ${block.uri || "unknown"}]`);
+				break;
+			default:
+				if (block?.text) parts.push(String(block.text));
+				break;
+		}
+	}
```
shreyas-lyzr, line 65: The `listAllTools` loop has no timeout per page fetch, only the outer `withTimeout` in `setupMcp` covers the whole `buildToolsForConnection` call. If a server stalls mid-pagination (returns the first page but hangs on the second), the 30-second timeout does still fire — just noting this relies on the outer guard rather than per-request timeouts. Fine as-is if that tradeoff is intentional.

src/mcp/manager.ts
```diff
@@ -0,0 +1,254 @@
...
+					return flattenToolResult(result);
+				} catch (err: any) {
+					// Protocol-level errors (e.g. -32602) throw; surface them as a
+					// string so the agent can read and recover instead of crashing.
+					return `Error: ${err?.message || err}`;
+				}
+			},
+			// We can't know a remote tool's semantics — fail closed.
+			metadata: { isConcurrencySafe: false, isReadOnly: false, isDestructive: false },
+		}),
```
shreyas-lyzr, line 136: The hardcoded version string `"1.5.2"` in the Client identity will drift from the actual package version. Since package.json is already read elsewhere in the codebase, consider importing the version from there (or from a generated constant) so it stays accurate without manual updates on every release.

test/mcp.test.ts
Nivesh353: Fixed.— res.content[0].text → res directly since execute() returns the plain string from flattenToolResult.

src/mcp/manager.ts
Nivesh353: Fixed. Replaced Promise.race with an explicit new Promise that calls clearTimeout(timer) in both the resolve and reject paths.

src/env-utils.ts
Nivesh353: Fixed. Switched to ?? "" and added console.warn when process.env[envName] === undefined, so a missing auth token logs [mcp] env var TOKEN is not set instead of silently becoming an empty string.

src/mcp/manager.ts
Nivesh353: Intentional — relying on the outer withTimeout in setupMcp as the single guard for the full buildToolsForConnection call. Adding per-request timeouts inside the pagination loop would add complexity without much gain since the 30s outer timeout still fires if any page hangs. Happy to revisit if you see a case where this falls short.

src/mcp/manager.ts
Nivesh353: Fixed. Imported version from package.json via createRequire so it stays in sync automatically on every release. The hardcoded value was already stale (2.0.0 in package.json).

shreyas-lyzr (approved):

- withTimeout now clears the timer on both the resolve and reject paths — no more handle leak.
- interpolateEnvString warns when an env var is unset instead of silently substituting an empty string.
- PACKAGE_VERSION is sourced from package.json via createRequire rather than being hardcoded.
- Test assertions correctly use assert.equal(res, ...) on the flattened string return, not .content[0].text.

The new test suite looks solid — flattenToolResult unit tests cover all block types including the structuredContent fallback, buildToolsForConnection tests cover pagination, name sanitization, protocol errors, and the in-memory integration path. Code is clean and the fail-soft design is well-considered. Good to merge.

PR #60, opened 2026-06-24
feat(a2a): add outbound A2A client (CLI + SDK)

Nivesh353:
## Summary

Adds an **A2A (Agent2Agent) client** so gitagent can call remote A2A-compatible
agents — built on LangGraph, CrewAI, Google ADK, or any A2A server — and use
their results mid-conversation. Each remote skill shows up as a normal tool, so
delegation looks identical to any other tool call.

**Outbound only — no server is started.** The feature is fully opt-in: with no
`a2a_agents` configured, behavior is unchanged and zero network calls are made.

> A2A is complementary to local tools: tools give the agent *capabilities*, A2A
> gives it *peers*.

## Motivation

Today gitagent can't talk to other agents. As teams build specialized agents on
other frameworks, users want gitagent to **delegate** to them instead of doing
everything itself. A2A is the open, Linux-Foundation interop standard for exactly
this, so it's the natural way to make gitagent interoperable.

## How it works

1. **Discovery** — at startup, for each configured agent gitagent fetches its
   **Agent Card** (`/.well-known/agent-card.json`).
2. **Tools** — each skill becomes one tool named `<agent>__<skill>`
   (skill-less agents → a single tool named after the agent). Name collisions
   are skipped with a warning.
3. **Delegation** — when the model calls one, gitagent sends the task via
   `message/send` (blocking) or `message/stream` (SSE streaming), flattens the
   reply to text, and returns it.

Failed/unreachable agents are **non-fatal** — they warn and are skipped; the rest
of the session works normally.

## Configuration

```yaml
# agent.yaml
a2a_agents:
  research-agent:
    url: https://research.example.com          # required
    headers:                                    # optional — ${VAR} read from env
      Authorization: "Bearer ${RESEARCH_TOKEN}"
    stream: true                                # optional (default true; false = blocking)
    timeoutMs: 30000                            # optional (default 30000)
    cardPath: /.well-known/agent-card.json      # optional override
```
## Testing
Unit — test/a2a.test.ts (14 cases): skill→tool mapping + namespacing, zero-skill fallback, name collisions, missing/unreachable agents (non-fatal), blocking, error propagation, streaming + partial updates. Full suite: 41/41 green.
End-to-end — verified against two real A2A servers built for this: a LangGraph agent with tools and a CrewAI multi-agent crew. Confirmed discovery, tool registration, blocking + streaming, bearer-token auth (correct/wrong/missing → connect/skip), and timeout handling — through both the CLI and the SDK.


src/sdk.ts
```diff
@@ -189,6 +196,17 @@ export function query(options: QueryOptions): Query {
...
 
+		// A2A remote-agent tools — outbound only, opt-in. Sourced from the
+		// manifest's `a2a_agents` and/or the `a2aAgents` option (option wins on
+		// key collision), so SDK users can define agents purely in code.
+		const a2aAgents =
+			loaded.manifest.a2a_agents || options.a2aAgents
+				? { ...loaded.manifest.a2a_agents, ...options.a2aAgents }
+				: undefined;
+		const a2aSetup = await setupA2A(a2aAgents, new Set(tools.map((t) => t.name)));
+		tools = [...tools, ...a2aSetup.tools];
```
shreyas-lyzr, line 210: Operator-precedence bug in the merge condition.

The intent is "pass undefined when neither source has agents", but the current expression:
```ts
loaded.manifest.a2a_agents || options.a2aAgents
    ? { ...loaded.manifest.a2a_agents, ...options.a2aAgents }
    : undefined;
```
is parsed as `(loaded.manifest.a2a_agents) || (options.a2aAgents ? … : undefined)` — not `(a || b) ? … : undefined`. When `loaded.manifest.a2a_agents` is a non-empty object (truthy), the ternary short-circuits to the spread regardless of `options.a2aAgents`, which is fine. But when `loaded.manifest.a2a_agents` is undefined/null and `options.a2aAgents` is also undefined, the expression evaluates `options.a2aAgents ? … : undefined` which also gives undefined — so the final value is accidentally correct. However, the logic is fragile and the precedence is wrong, and this will silently misbehave if the ternary branches are ever reordered. Add explicit parens:
```suggestion
const a2aAgents =
    (loaded.manifest.a2a_agents || options.a2aAgents)
        ? { ...loaded.manifest.a2a_agents, ...options.a2aAgents }
        : undefined;
```

shreyas-lyzr (commented):
Overall the implementation is solid — clean separation into manager/types, good non-fatal error handling, reasonable test coverage. Three issues worth addressing: one correctness bug in the SDK merge logic, one architecture concern about module-scoped cleanup state, and one minor safety gap in the cleanup chain. Details inline.

src/index.ts
```diff
@@ -40,6 +41,18 @@ const bold = (s: string) => `\x1b[1m${s}\x1b[0m`;
...
 
+// Teardown for A2A connections, set during agent assembly. Module-scoped so the
+// SIGTERM handler (outside main()) can reach it. No-op until A2A is configured.
+let a2aCleanup: (() => Promise<void>) | null = null;
+const runA2ACleanup = async () => {
+	if (!a2aCleanup) return;
+	try {
+		await a2aCleanup();
+	} catch {
+		/* best-effort — never block exit */
```
shreyas-lyzr, line 55: Module-scoped mutable cleanup state is fragile.

`a2aCleanup` is a module-level `let`. In the CLI this is fine (one session, then exit). But if `main()` is ever called more than once in the same process (test harness, watch mode, SDK reuse), the second invocation overwrites the first session's cleanup reference, causing a silent leak. The SDK path correctly handles this with a local variable and closure — the same pattern would be safer here too.

At minimum, reset `a2aCleanup = null` after `runA2ACleanup()` is called, so accidental double-cleanup is safe and a fresh session always starts clean.

src/index.ts
```diff
@@ -836,7 +857,7 @@ async function main(): Promise<void> {
 			try {
 				_session.end({ "gitagent.cost_usd": _totalCostUsd });
 			} catch { /* ignore */ }
-			stopSandbox().finally(() => process.exit(0));
+			runA2ACleanup().finally(() => stopSandbox().finally(() => process.exit(0)));
 		}
```
shreyas-lyzr, line 872: Cleanup fires before `stopSandbox` on SIGINT (via `.finally`), but not before `shutdownTelemetry` on SIGTERM.

The SIGTERM handler is:
```ts
runA2ACleanup()
    .finally(() => shutdownTelemetry().catch(() => {}))
    .finally(() => process.exit(0));
```
A2A cleanup runs first — good. But the SIGINT handler added in this PR chains as:
```ts
runA2ACleanup().finally(() => stopSandbox().finally(() => process.exit(0)));
```
Telemetry shutdown is skipped entirely on SIGINT. That was true before this PR (pre-existing issue), but since the shutdown chain is being touched here anyway, it is worth making the two signal handlers consistent. Not blocking, but worth flagging.

src/sdk.ts
Nivesh353: Fixed. Added explicit parens: (loaded.manifest.a2a_agents || options.a2aAgents) ? … : undefined. The logic was accidentally correct today but would silently break if the branches were reordered. Updated in src/sdk.ts line 202.

src/index.ts
Nivesh353: Fixed. runA2ACleanup now captures the ref into a local fn, nulls out a2aCleanup before calling — so double-cleanup is safe and a fresh main() invocation always starts with a clean slate. Updated in src/index.ts lines 47–54.

src/index.ts
Nivesh353: Fixed. SIGINT now chains runA2ACleanup → shutdownTelemetry → stopSandbox → process.exit, matching the SIGTERM handler exactly. Updated in src/index.ts line 860.

shreyas-lyzr (approved):
All three comments addressed:

- Operator-precedence fix: the `a2aAgents` merge now correctly uses `{ ...loaded.manifest.a2a_agents, ...options.a2aAgents }` guarded by the truthiness check, so the merge is unambiguous.
- Module-scoped `a2aCleanup` in `src/index.ts` is now nulled out before invoking (`const fn = a2aCleanup; a2aCleanup = null; await fn()`), preventing double-execution in repeated-call scenarios.
- SIGTERM handler now chains `runA2ACleanup()` before telemetry shutdown, aligning the two exit paths.

Good to merge.

PR #62, opened 2026-06-26
fix(tool-loader): enable image output from declarative tools via data URI

Nivesh353:
## Problem

Declarative YAML tools that return image data (e.g. a custom `read_image` tool) were broken silently. `tool-loader.ts` always wrapped tool output as `type: "text"`, so the model received raw base64 characters as plain text instead of an image content block. The tool ran without errors, making it impossible for users to know why vision analysis wasn't working — the model would hallucinate a description instead.

## Fix

Added data URI detection in `tool-loader.ts`. When a tool's script outputs a line in the format:

data:image/jpeg;base64,<base64data>

The loader parses the mime type and base64 data from the URI and returns a proper `type: "image"` content block to the model. Internal whitespace is stripped to handle scripts that don't normalize base64 output.

## Why Data URI

Avoids any field naming convention — users can name their script variables anything they want. The data URI format is a web standard and self-describing. No changes needed to the YAML tool definition.

## Provider Support

Works across all providers. `pi-ai`'s shared `transform-messages` layer handles image content uniformly — vision-capable models receive the actual image, non-vision models automatically receive `"(tool image omitted: model does not support images)"` with no extra handling required.

## Testing

Tested with a custom `read_image` declarative tool on a JPEG file using a vision-capable model. Model correctly described the image contents after this fix.

### DIFF
--- a/src/tool-loader.ts
+++ b/src/tool-loader.ts
@@ -118,6 +118,19 @@ function createDeclarativeTool(
 
 					// Try parsing JSON output
 					let text = stdout.trim();
+
+					// Detect data URI — works regardless of field naming in user scripts
+					if (text.startsWith("data:image/") && text.includes(";base64,")) {
+						const commaIndex = text.indexOf(",");
+						const mimeType = text.slice(5, commaIndex).split(";")[0];
+						const data = text.slice(commaIndex + 1).replace(/\s/g, "");
+						resolve({
+							content: [{ type: "image", data, mimeType }],
+							details: undefined,
+						});
+						return;
+					}
+
 					try {
 						const parsed = JSON.parse(text);
 						if (parsed.text) text = parsed.text;
--- a/src/tool-loader.ts
+++ b/src/tool-loader.ts
@@ -119,16 +119,25 @@ function createDeclarativeTool(
 					// Try parsing JSON output
 					let text = stdout.trim();
 
-					// Detect data URI — works regardless of field naming in user scripts
-					if (text.startsWith("data:image/") && text.includes(";base64,")) {
-						const commaIndex = text.indexOf(",");
-						const mimeType = text.slice(5, commaIndex).split(";")[0];
-						const data = text.slice(commaIndex + 1).replace(/\s/g, "");
-						resolve({
-							content: [{ type: "image", data, mimeType }],
-							details: undefined,
-						});
-						return;
+					// Detect data URI — works regardless of field naming in user scripts.
+					// Scan lines so stray log output before the URI doesn't break detection.
+					const dataUriLine = text.split("\n").find(
+						line => line.startsWith("data:image/") && line.includes(";base64,"),
+					);
+					if (dataUriLine) {
+						const commaIndex = dataUriLine.indexOf(",");
+						// Slice past the leading "data:" prefix (5 chars) to get "image/jpeg;base64"
+						const mimeType = dataUriLine.slice(5, commaIndex).split(";")[0];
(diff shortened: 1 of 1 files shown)


src/tool-loader.ts
```diff
@@ -118,6 +118,19 @@ function createDeclarativeTool(
 
 					// Try parsing JSON output
 					let text = stdout.trim();
+
```
shreyas-lyzr, line 121: The slice is correct, but the magic number `5` isn't obvious. A comment here would help:
```suggestion
					// Slice past the leading "data:" prefix (5 chars) to get "image/jpeg;base64"
					const mimeType = text.slice(5, commaIndex).split(";")[0];
```

src/tool-loader.ts
```diff
@@ -118,6 +118,19 @@ function createDeclarativeTool(
...
+					// Detect data URI — works regardless of field naming in user scripts
```
shreyas-lyzr: Consider guarding against an empty data string and an unexpected mime type:

```suggestion
					const data = text.slice(commaIndex + 1).replace(/\s/g, "");
					const allowedMimeTypes = ["image/jpeg", "image/png", "image/gif", "image/webp"];
					if (!data || !allowedMimeTypes.includes(mimeType)) {
						// Fall through to text/JSON handling rather than silently sending bad content
					} else {
```

SVG is technically an image MIME type but is not a raster format; several provider vision APIs reject it or behave unexpectedly.

shreyas-lyzr (commented):
Small, focused patch — the motivation is solid and the approach is clean. A few things worth addressing before merge.

**mimeType extraction** (line 122): The slice logic is correct but subtle. `text.slice(5, commaIndex).split(";")[0]` works, but a brief comment explaining the `5` (skipping "data:") would help future readers.

**No validation of the base64 payload**: After stripping whitespace, `data` is passed directly to the content block with no checks. If a script outputs a truncated URI, empty base64, or an unsupported media type like `data:image/svg+xml` (which some provider backends handle differently from raster formats), the failure is silent — the same problem the fix is trying to avoid. Consider rejecting empty `data` and optionally allowlisting known-safe mime types (jpeg, png, gif, webp).

**Single-line assumption**: Detection relies on `text.startsWith("data:image/")`. If the script writes any output before the data URI — a stray log line, a warning — this silently falls through to JSON parsing. A more robust approach would scan for the first line matching `data:image/` rather than requiring the entire stdout to be the URI. Real scripts leak to stdout more often than expected.

**No automated test**: The PR is manually tested. Given this is a content-routing path that determines what the model actually receives, a regression test covering the data URI branch would be valuable. Not blocking if the project doesn't have test infrastructure here, but worth a follow-up.

src/tool-loader.ts
Nivesh353: Added the comment as suggested — // Slice past the leading "data:" prefix (5 chars) to get "image/jpeg;base64". Done.

src/tool-loader.ts
Nivesh353: Added both guards. allowedMimeTypes allowlists jpeg, png, gif, webp. Empty data check prevents silently sending a truncated URI. Both fall through to text/JSON handling rather than rejecting with an error. SVG and other non-raster types are excluded as suggested.

shreyas-lyzr (approved):
All four concerns from the previous review are addressed: line-scan handles leading log output, the slice comment is in place, empty data is guarded, and the allowed MIME type list blocks non-raster types. Good to merge.

PR #66, opened 2026-07-02
feat: doc-converter + CCR — PDF/DOCX/XLSX/PPTX reading and large doc token efficiency

### LINKED ISSUE #65: feat: doc-converter + CCR — PDF/DOCX/XLSX/PPTX reading and large doc token efficiency
## Problem

gitagent currently cannot read PDF, DOCX, XLSX, or PPTX files. Clients whose users upload documents get a `[Binary file]` response — the agent is blind to those files entirely.

Additionally, when agents work with large documents across multiple turns, the full document content accumulates in context on every turn, compounding API costs with each message.

## Proposed Solution

A two-part pipeline built into the `read` tool:

### 1. Doc-converter — binary → readable
Convert PDF, DOCX, XLSX, PPTX to clean markdown automatically when `read()` is called. No manual pre-processing needed by the client.

- Disk cache at `memory/.doc-cache/` with mtime-based invalidation
- Soft cap: files >50MB or >500 pages return a warning instead of stalling silently
- If conversion fails, surface an explicit error to the agent (no silent lossy fallback)
- Known edge case: same mtime + different inode won't invalidate cache — acceptable for v1, noted in code

### 2. CCR (Chunked Reversible Retrieval) — large doc efficiency
For documents exceeding a token threshold, return a lightweight outline on first read. The agent fetches only the sections it needs on demand.

**Locked API surface:**
```
read_doc_section(doc_id, section_id) → { content, tokens, has_refs: [section_id, ...] }
```
`has_refs` returns a list of section IDs referenced in the fetched content — lets the agent decide in one hop whether to follow up, without re-parsing the text itself.

**Dry-run mode (v1 scope):**
```
read(file, mode="outline") → { outline, tokens }
```
Returns outline + token estimates without fetching content. Lets the agent make an informed fetch decision before committing.

**Section boundaries:** markdown headings as chunk boundaries, flat heuristics for PDFs without heading structure. Token estimates are pre-computed at chunk time.

**DocStore:** in-memory per session, GC'd on session end. Soft LRU cap to avoid OOM on sessions reading many large docs.

## Scope (v1)
- `src/tools/read.ts` — pipeline wired in, `mode="outline"` support
- `src/tools/doc-converter.ts` — new (PDF/DOCX/XLSX/PPTX → markdown, file size guard, explicit error on failure)
- `src/tools/doc-store.ts` — new (CCR + `read_doc_section` with `has_refs`, LRU soft cap)

## Known limitations (v2)
- Cross-section dependency resolution — agent's responsibility for now
- Streaming / incremental conversion for very large PDFs
- Multi-user persistence across sessions
- SmartCrusher, CodeCompressor, CacheAligner — tracked separately

## Benchmark methodology
Compare two **successful** agent runs on the same document:
- **Baseline** — doc-converter converts PDF, full markdown injected into prompt
- **With CCR** — agent receives outline, fetches only relevant sections on demand

This gives a fair apples-to-apples token comparison where both runs complete the task.

stealthwhizz:
## Summary

- **Doc-converter** — gitagent can now read PDF, DOCX, XLSX, PPTX files. Converts to clean markdown automatically on `read()`. Previously returned `[Binary file]`.
- **CCR (Chunked Reversible Retrieval)** — documents above 8K tokens return a lightweight outline first. Agent fetches only the sections it needs via `read_doc_section`, preventing large docs from flooding context on every turn.
- **Benchmark script** — `test-live-benchmark.mjs` compares two successful runs: full markdown in prompt (no CCR) vs outline + fetch (with CCR). Supports Anthropic, OpenAI, and Lyzr.

## What changed

| File | Change |
|---|---|
| `src/tools/doc-converter.ts` | New — PDF/DOCX/XLSX/PPTX → markdown, 50MB file size guard, explicit errors |
| `src/tools/doc-store.ts` | New — CCR chunk store, `read_doc_section` with `has_refs`, LRU soft cap (10MB) |
| `src/tools/read.ts` | Modified — doc-converter + CCR pipeline wired in, `mode="outline"` dry-run |
| `src/tools/shared.ts` | Modified — `mode="outline"` added to read schema |
| `src/tools/index.ts` | Modified — `read_doc_section` tool registered with `has_refs` response |
| `test-live-benchmark.mjs` | New — live before/after benchmark, multi-provider |

## API surface (locked)

```
read(path, mode="outline") → outline + token estimates (dry-run, no content loaded)
read_doc_section(doc_id, section_id) → { content, tokens, has_refs: [section_id, ...] }
```

`has_refs` lets the agent decide follow-up fetches in one hop without re-parsing content.

## Known limitations (v2)
- Cross-section dependency resolution — agent's responsibility for now
- Streaming / incremental conversion for very large PDFs
- Multi-user persistence across sessions
- SmartCrusher, CodeCompressor, CacheAligner — tracked in issue #65

Closes #65

### DIFF
--- a/src/tools/doc-converter.ts
+++ b/src/tools/doc-converter.ts
@@ -13,10 +13,11 @@ export interface ConversionResult {
 	savedPercent: number;
 }
 
+// .doc and .ppt are legacy binary formats — mammoth/JSZip can't reliably parse them.
+// Only the modern XML-based formats (.docx, .xlsx, .pptx) are supported.
 const CONVERTIBLE = new Set([
-	".pdf", ".docx", ".doc",
-	".xlsx", ".xls",
-	".pptx", ".ppt",
+	".pdf",
+	".docx", ".xlsx", ".pptx",
 	".html", ".htm",
 ]);
 
--- a/src/tools/doc-converter.ts
+++ b/src/tools/doc-converter.ts
@@ -1,0 +1,231 @@
+import { extname, basename } from "path";
+import { estimateTokens } from "../compact.js";
+import { PDFParse } from "pdf-parse";
+import mammoth from "mammoth";
+import * as XLSX from "xlsx";
+import JSZip from "jszip";
+
+export interface ConversionResult {
+	markdown: string;
+	originalTokens: number;
+	convertedTokens: number;
+	savedTokens: number;
+	savedPercent: number;
+}
+
+const CONVERTIBLE = new Set([
+	".pdf", ".docx", ".doc",
+	".xlsx", ".xls",
+	".pptx", ".ppt",
+	".html", ".htm",
+]);
+
+export function isConvertible(filePath: string): boolean {
+	return CONVERTIBLE.has(extname(filePath).toLowerCase());
+}
+
+// ── Format converters ──────────────────────────────────────────────────
+
+async function pdfToMarkdown(buffer: Buffer): Promise<string> {
+	const parser = new PDFParse({ data: buffer });
+	const result = await parser.getText();
+	return result.text.trim();
+}
+
+async function docxToMarkdown(buffer: Buffer): Promise<string> {
+	const result = await mammoth.convertToHtml({ buffer });
+	return htmlToText(result.value);
+}
+
+async function xlsxToMarkdown(buffer: Buffer): Promise<string> {
+	const workbook = XLSX.read(buffer, { type: "buffer" });
+	const sections: string[] = [];
+
+	for (const sheetName of workbook.SheetNames) {
+		const sheet = workbook.Sheets[sheetName];
+		const rows = XLSX.utils.sheet_to_json<unknown[]>(sheet, { header: 1, defval: "" });
+		if (rows.length === 0) continue;
+
+		const [header, ...body] = rows as string[][];
+		const separator = header.map(() => "---");
+		const table = [header, separator, ...body]
+			.map((row) => "| " + row.map(String).join(" | ") + " |")
+			.join("\n");
+
+		sections.push(`## ${sheetName}\n\n${table}`);
+	}
+
+	return sections.join("\n\n");
+}
+
+async function pptxToMarkdown(buffer: Buffer): Promise<string> {
+	const zip = await JSZip.loadAsync(buffer);
+	const slideEntries = Object.keys(zip.files)
+		.filter((name) => /^ppt\/slides\/slide\d+\.xml$/.test(name))
+		.sort((a, b) => {
+			const n = (s: string) => parseInt(s.match(/\d+/)![0], 10);
+			return n(a) - n(b);
+		});
+
+	const slides: string[] = [];
+	for (let i = 0; i < slideEntries.length; i++) {
+		const xml = await zip.files[slideEntries[i]].async("string");
+		const texts: string[] = [];
+		for (const match of xml.matchAll(/<a:t[^>]*>([^<]*)<\/a:t>/g)) {
+			const text = match[1].trim();
+			if (text) texts.push(text);
+		}
+		if (texts.length > 0) {
+			slides.push(`## Slide ${i + 1}\n\n${texts.join("\n\n")}`);
+		}
+	}
+
+	return slides.join("\n\n---\n\n");
+}
+
+function htmlToText(html: string): string {
+	return html
+		.replace(/<style[^>]*>[\s\S]*?<\/style>/gi, "")
+		.replace(/<script[^>]*>[\s\S]*?<\/script>/gi, "")
+		.replace(/<br\s*\/?>/gi, "\n")
+		.replace(/<\/p>/gi, "\n\n")
+		.replace(/<\/h[1-6]>/gi, "\n\n")
+		.replace(/<[^>]+>/g, "")
+		.replace(/&nbsp;/g, " ")
+		.replace(/&amp;/g, "&")
+		.replace(/&lt;/g, "<")
+		.replace(/&gt;/g, ">")
+		.replace(/&quot;/g, '"')
+		.replace(/\n{3,}/g, "\n\n")
+		.trim();
+}
+
+// ── Post-processing (Stage 2a — headroom-style compression) ───────────
+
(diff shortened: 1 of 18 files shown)


src/tools/doc-converter.ts
```diff
@@ -0,0 +1,242 @@
+import { extname, basename } from "path";
+import { estimateTokens } from "../compact.js";
+import { PDFParse } from "pdf-parse";
```
shreyas-lyzr, line 3: The import `{ PDFParse }` is wrong for pdf-parse v2. The package's default export is a plain async function, not a named class. This will throw `TypeError: PDFParse is not a constructor` at runtime on any PDF read.

The correct usage is:
```suggestion
import pdfParse from "pdf-parse";
```
And in `pdfToMarkdown`:
```ts
async function pdfToMarkdown(buffer: Buffer): Promise<string> {
	const result = await pdfParse(buffer);
	return result.text.trim();
}
```
Worth adding a quick smoke test (even just `node -e "require('pdf-parse')(Buffer.alloc(0)).catch(() => {})"`) to catch API mismatches like this before CI.

src/tools/read.ts
```diff
@@ -20,29 +28,145 @@ function isBinary(buffer: Buffer): boolean {
 	return false;
 }
 
-export function createReadTool(cwd: string): AgentTool<typeof readSchema> {
+function estimateTokens(s: string): number {
+	return Math.ceil(s.length / 4);
+}
+
+function cacheKey(absolutePath: string): string {
```
shreyas-lyzr, line 35: `cacheKey` replaces `/`, `\\`, and `:` with `_`, which causes path collisions. For example:

- `/tmp/a/b.pdf` and `/tmp/a_b.pdf` both produce `_tmp_a_b.pdf.md`

A stale or wrong cached conversion will be silently served with no error. Use a collision-resistant key — a hash of the absolute path, or at minimum double the separator:

```suggestion
function cacheKey(absolutePath: string): string {
	// Use a hash to avoid collisions from path separator substitution
	const hash = absolutePath.split("").reduce((h, c) => (Math.imul(31, h) + c.charCodeAt(0)) | 0, 0);
	const safe = absolutePath.replace(/[/\\:]/g, "_").slice(-60);
	return `${Math.abs(hash).toString(36)}_${safe}.md`;
}
```

src/tools/doc-store.ts
```diff
@@ -0,0 +1,195 @@
...
+	fetch(filePath: string, chunkId: string): FetchResult | null {
+		const doc = this.docs.get(filePath);
+		if (!doc) return null;
+		const chunk = doc.chunks.find((c) => c.id === chunkId);
+		if (!chunk) return null;
+
+		doc.lastAccessed = Date.now();
+
+		// Extract section IDs referenced in this chunk's content (e.g. "see s3" or "[s3]")
+		const has_refs = [...chunk.content.matchAll(/\[?(s\d+)\]?/g)]
```
shreyas-lyzr: The `has_refs` regex `/\[?(s\d+)\]?/g` matches any substring of the form `sN` — including words like `s3ction`, `s10me`, or `session3`. Running this against real document text will produce noisy false-positive refs that lead the agent to fetch unrelated sections.

Tighten the match to whole-word / bracket-delimited occurrences:

```suggestion
const has_refs = [...chunk.content.matchAll(/(?<![a-zA-Z])\[s(\d+)\]|(?<![a-zA-Z])s(\d+)(?!\w)/g)]
	.map((m) => m[1] ?? m[2])
	.filter((id) => {
		const sectionId = `s${id}`;
		return sectionId !== chunkId && doc.chunks.some((c) => c.id === sectionId);
	});
```
Or, simpler and safer for v1: only match the explicit bracket form `[sN]` which is what the outline output uses, and skip the bare-word match entirely.

src/tools/index.ts
```diff
@@ -41,13 +45,51 @@ export function createBuiltinTools(config: BuiltinToolsConfig): AgentTool<any>[]
...
+				type: "object",
+				properties: {
+					path: { type: "string", description: "Absolute or relative path to the document (same as passed to read)" },
+					section_id: { type: "string", description: "Section ID from the outline (e.g. s1, s4)" },
+				},
+				required: ["path", "section_id"],
+			},
+			execute: async (_toolCallId: string, params: unknown) => {
+				const { path, section_id } = params as { path: string; section_id: string };
+				const altPath = path.startsWith("/") ? path : `${config.dir}/${path}`.replace(/\\/g, "/");
```
shreyas-lyzr: The `altPath` construction uses string concatenation instead of `path.resolve()`, which can produce a path like `/cwd/./docs/report.pdf` when the user passes `./docs/report.pdf`. Meanwhile `read.ts` stores the key using `resolve(cwd, path)` which normalizes to `/cwd/docs/report.pdf`. These won't match in the Map, so `docStore.fetch(path, section_id)` and `docStore.fetch(altPath, section_id)` both return null and the agent gets `[No section found]`.

```suggestion
const altPath = resolve(config.dir, path);
```
(Import `resolve` from `"path"` at the top of `index.ts`.)  This makes the lookup key consistent with what `read.ts` stores.

shreyas-lyzr (changes requested):
Solid implementation — the doc-converter + CCR pipeline is well-structured and matches the agreed spec. Four issues below that need fixing before merge: one is a runtime crash, one is a key-collision bug, one is a section-ID false-positive, and one is a path mismatch that will silently break read_doc_section on relative paths. Everything else is in good shape.

shreyas-lyzr (approved):
All four issues from the previous review are resolved:

1. PDFParse import — pdf-parse v2.4.5 exports PDFParse as a named class, so the import and constructor call are correct for the pinned version. The previous concern was based on v1 behavior; no longer applies.

2. Cache key collision — fixed with a djb2 hash appended to the path. /tmp/a/b.pdf and /tmp/a_b.pdf now produce distinct keys.

3. has_refs false positives — fixed by requiring bracket syntax [sN] and filtering against actual chunk IDs from the store.

4. Path mismatch in read_doc_section — fixed by using resolve(config.dir, path) for altPath and trying both the raw and resolved paths.

Implementation matches the agreed spec. Good to merge.

PR #69, opened 2026-07-06
fix: bound the tool-use loop to prevent runaway provider/tool cycles

### LINKED ISSUE #58: GitAgent can duplicate tool execution under repeated tool-use state
# GitAgent can duplicate tool execution under repeated tool-use state

## Summary

Severity / performance impact: High. A repeated tool-use state can lead to a high-volume provider/tool-result loop in a repository workspace.

## Affected target

- Target: `GitAgent`
- Version/build: `@open-gitagent/gitagent 2.0.0`
- Environment: Linux x86_64 Docker, public package build
- Provider/model: OpenAI-compatible local mock provider, `gpt-4`

## Steps to reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It builds the affected CLI version from the public package registry, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/f63d9147d36cbce66937e6602068e2a0

The Gist contains `gitagent-empty-write-duplicate-execution.reproduce.py`. Download and run:

```bash
gh gist clone f63d9147d36cbce66937e6602068e2a0 gitagent-empty-write-duplicate-execution-reproducer
cd gitagent-empty-write-duplicate-execution-reproducer
python3 gitagent-empty-write-duplicate-execution.reproduce.py
```

To reuse an already-built local image:

```bash
python3 gitagent-empty-write-duplicate-execution.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```text
REPRODUCED
```

## Actual behavior

The local provider repeats a tool call after tool results arrive. The script checks for at least 100 provider requests, duplicated tool results, and timeout/signal termination.

## Expected behavior

GitAgent should bound repeated tool-use state and avoid repeatedly resubmitting the same tool result.

## Public-upstream status

This report targets the stock public `@open-gitagent/gitagent@2.0.0` package. A complete self-contained reproducer is linked above and builds the public package in an isolated Docker workspace.

stealthwhizz:
## Problem

pi-agent-core's agent loop continues for as long as the assistant keeps emitting tool calls — there is no turn cap in the loop, and the `maxTurns` value gitagent sets on the Agent state is never consumed by the library.

A provider that re-requests the same tool call after every tool result (a broken or malicious backend) drives an **unbounded loop**: 100+ provider requests, duplicated tool results, and eventual timeout. This is the DoS-class behavior in #58.

## Root cause

- `runLoop` in pi-agent-core loops on `while (hasMoreToolCalls || pendingMessages.length > 0)` with `hasMoreToolCalls = !executedToolBatch.terminate`. Nothing counts turns.
- `grep maxTurns node_modules/@mariozechner/pi-agent-core` → no matches. gitagent's `modelOptions.maxTurns` is spread into `initialState` but never read.

## Fix

A loop guard (`src/loop-guard.ts`) wired through the `afterToolCall` hook, which is the library's supported termination path (`terminate: true` on every result in a batch stops the loop). It terminates the run when either bound is hit:

- **maxTurns** — total assistant turns that request tools. Precedence: explicit `query()` option → agent manifest `runtime.max_turns` → default `50`.
- **maxRepeats** — 5 consecutive identical tool calls (same name + args), the signature of a stuck loop. Catches the reproducer well under the 100-request threshold.

The guard latches once tripped so a mid-batch trip still terminates the next batch, and it preserves the last tool's real output while appending the stop reason for observability.

## Tests

`test/loop-guard.test.ts` — normal varied use (no trip), repeat detection, counter reset on arg change, maxTurns cap, per-batch turn counting (shared assistantMessage reference = one turn), and output preservation. Full suite: 33 passing.

Closes #58


shreyas-lyzr (approved):
Solid fix for the unbounded loop issue (#58). The implementation is clean and the test coverage is thorough.

One minor observation: `maxRepeats` is clamped to `Math.max(2, ...)`, meaning a caller passing `maxRepeats: 1` gets silently bumped to 2. This is a reasonable guard but worth a comment noting that values below 2 are unsupported — not a blocker.

Everything else checks out: the `afterToolCall` hook is the correct termination path, the latch-once behavior ensures a mid-batch trip propagates correctly, per-batch turn counting via reference equality is the right approach, and the precedence order for `maxTurns` (explicit option → manifest → default) is sensible. All six test cases cover the meaningful edge cases.

shreyas-lyzr (approved):
Solid, targeted fix for the unbounded tool-use loop (issue #58).

**What the PR does:** Introduces a `LoopGuard` wired through pi-agent-core's `afterToolCall` hook — the library's supported termination path. When either bound is hit, the guard returns `terminate: true` on every result in the batch, which stops the loop cleanly. Two independent bounds:
- `maxTurns`: total assistant turns that invoked tools. Defaults to 50, can be overridden by `manifest.runtime.max_turns` or an explicit query option.
- `maxRepeats`: 5 consecutive identical tool calls (same name + args), catching stuck loops well before any turn cap.

**Implementation review:**

`loop-guard.ts`: The latching pattern (`trippedReason`) is critical — without it, a mid-batch trip would only terminate some results in the batch, not all, so the loop wouldn't stop. The latch correctly propagates through the rest of that batch and all subsequent batches. Good.

Turn counting by `assistantMessage` object reference is the right mechanism here — all tool calls in one batch share the same reference, so only distinct batches increment the counter. The test that verifies this (`shared assistantMessage reference = one turn`) directly validates the behavior.

`signatureOf()` uses `JSON.stringify` with a fallback to `String(args)` — covers the `undefined`/non-serializable edge cases cleanly.

Output preservation: the guard prepends the real `content` and `details` from the tool result and appends only the stop-reason text. Callers that inspect tool results still get the real output.

`sdk.ts`: The `maxTurns` precedence chain (explicit option → manifest → 50) is correct and the value is reflected in `modelOptions.maxTurns` for telemetry/inspection. The guard itself does the actual enforcement.

**Tests (`loop-guard.test.ts`):** Covers all meaningful paths: normal use (no trip), repeat detection, counter reset on arg change, maxTurns cap, per-batch turn counting via shared reference, and output preservation on termination. Test coverage is thorough for the changed logic.

shreyas-lyzr (approved):
Loop guard looks solid. A few notes worth calling out, none blocking.

The binary diff in the GitHub UI for loop-guard.ts (showing as 'Binary files /dev/null and b/src/loop-guard.ts differ') is a cosmetic git artifact from how the branch was created — the file is valid UTF-8 TypeScript when read directly. Not a problem with the code.

Logic correctness looks good: turn counting by assistantMessage reference identity correctly groups parallel tool calls in one batch as one turn, the repeat counter resets on any arg change, and the latch (trippedReason) ensures a mid-batch trip still terminates the following batch via terminate: true on every subsequent result.

One small edge case: maxRepeats clamps to Math.max(2, ...) even when the caller explicitly passes maxRepeats: 1. A caller who passes 1 expecting 'terminate immediately on first repeat' gets 2 instead, silently. This matters if the reproducer from #58 is tested with maxRepeats: 1 in a unit test — it would still pass because the test uses maxRepeats: 5, but the behaviour diverges from the option's documented semantics. Not blocking but worth documenting or relaxing the clamp to Math.max(1, ...).

The DEFAULT_MAX_TURNS = 50 constant is defined inline in sdk.ts rather than imported from loop-guard.ts where it arguably belongs. Minor cohesion point — the guard and its defaults should ideally live together.

Tests are thorough: they cover the normal path, repeat detection, counter reset, maxTurns cap, per-batch turn semantics, and output preservation. Good coverage. 6 test cases for a ~100-line module is the right ratio.

Security pass: no new dependencies, no hard-coded secrets, no injection sinks. Clean.

PR #70, opened 2026-07-06
fix: use manifest model.fallback at runtime for provider resilience

### LINKED ISSUE #24: Agent ignores configured files and returns credit balance error on queries
When running GitClaw on localhost:3333, generic conversational queries work correctly and receive valid responses. However, when asking questions that should be answered from the configured files (files and folders set up in the project), the agent fails to reference those files and instead returns a credit balance error.

**Steps to Reproduce**

- Configure both OpenAI (ChatGPT) and Anthropic (Claude) API keys in the project
- Start the agent - accessible at localhost:3333
- Ask a generic question (e.g. "Hello, how are you?") - works correctly
- Ask a question that should be answered from the knowledge base (e.g. "What is the leave policy?" or "What files have you been configured with?") fails

**Expected Behavior**

The agent should search the configured files and folders, retrieve relevant context, and return an accurate answer.

Image

stealthwhizz:
## Problem

The agent manifest supports a `model.fallback` list, and the scaffolding templates (`index.ts`, `session.ts`) tell users to fill it in — but **nothing ever read it at runtime**. `loadAgent` resolved only `model.preferred` and built a single-model agent.

So when the primary provider failed — e.g. Anthropic returning `credit balance too low` — the run died even though the user had configured a second provider and key expecting resilience. From the outside the agent looked like it "ignored the configured files" and returned a credit error (issue #24: generic chat works, file/knowledge queries die on the out-of-credits provider).

## Root cause

`grep -rn "model.fallback\|\.fallback" src/` → only the type declaration and scaffolding strings. No runtime consumer. `manifest.model.fallback` was dead config.

## Fix

**Resolve fallbacks (`loader.ts`)**
- Extract `resolveModel()` (custom endpoints, base-url override, key normalization) so the preferred model and every fallback entry resolve identically.
- Resolve `manifest.model.fallback` into an ordered `fallbackModels` list on `LoadedAgent`. Duplicate and unparseable entries are skipped rather than failing startup.

**Retry at runtime (`sdk.ts`)**
- Candidate list = `[preferred, ...fallbacks]`. Run the prompt on the current model; if it fails with a **retryable** provider error **before any output was streamed**, swap the model on the *same* agent (preserving conversation history), roll back the failed turn, and retry the next candidate.
- Each switch emits a `fallback` system message; only when all candidates are exhausted is the error surfaced. The errored assistant message is still emitted on terminal failure (preserves the existing `messages()` contract).
- Captures both failure paths: in-stream error (`message_end` with `stopReason: "error"`) and thrown errors routed through pi-agent-core's `handleRunFailure` (`agent_end` only).

**Classify errors (`model-fallback.ts`)**
- `isRetryableProviderError()` treats billing / credit / quota / rate-limit / auth / availability errors as retryable; request-shaped errors (bad `max_tokens`, content filter) are not — switching models won't fix those.

## Result for the #24 reporter

With OpenAI + Anthropic both configured and listed in `fallback`, a "leave policy" query where Anthropic is out of credits now transparently retries on OpenAI and answers from the files — no error. If *all* providers fail, they get one clear message naming what failed, instead of an agent that silently "ignores files."

## Tests

`test/model-fallback.test.ts` — retry classifier (billing/availability/auth → retry; empty → retry; request-shaped → no retry) and manifest fallback resolution (dedupe + skip blanks; empty list when unconfigured). Full suite: 33 passing.

Closes #24

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -385,9 +385,11 @@ export function query(options: QueryOptions): Query {
 							provider: msg.provider,
 							api: (msg as any).api,
 						};
-						// Reset accumulators and skip cost tracking (usage is empty on error).
+						// Reset accumulators + LLM-call timer (a partial stream may have
+						// set it) and skip cost tracking (usage is empty on error).
 						accText = "";
 						accThinking = "";
+						_llmCallStart = 0;
 						break;
 					}
 
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -23,6 +23,7 @@ import type {
 	SandboxOptions,
 } from "./sdk-types.js";
 import { CostTracker } from "./cost-tracker.js";
+import { isRetryableProviderError } from "./model-fallback.js";
 import { context as otelContext } from "@opentelemetry/api";
 import {
 	wrapToolWithOtel,
@@ -298,26 +299,45 @@ export function query(options: QueryOptions): Query {
 			modelOptions.maxTurns = options.maxTurns;
 		}
 
-		// 8. Create Agent
+		// 8. Fallback-aware agent runner.
+		// The manifest's preferred model plus any resolved fallbacks form an
+		// ordered candidate list. When a model fails with a retryable provider
+		// error before producing output (e.g. "credit balance too low"), the
+		// runner swaps the model on the same agent and retries (issue #24).
+		const candidateModels = [loaded.model, ...loaded.fallbackModels];
+
+		// Controller state shared across attempts.
+		let sessionStartEmitted = false;
+		interface AttemptState {
+			producedOutput: boolean;
+			error: { message?: string; provider?: string; model?: string; api?: string } | null;
+		}
+		let attempt: AttemptState = { producedOutput: false, error: null };
+
+		// 9. Build an Agent for a given model and wire event → GCMessage mapping.
+		const buildAgent = (model: typeof loaded.model): Agent => {
 		const agent = new Agent({
 			initialState: {
 				systemPrompt,
-				model: loaded.model,
+				model,
 				tools,
 				...modelOptions,
 			},
 		});
 
-		// 9. Subscribe to events and map to GCMessage
 		agent.subscribe((event: AgentEvent) => {
 			switch (event.type) {
 				case "agent_start":
-					pushMsg({
-						type: "system",
-						subtype: "session_start",
-						content: `Agent ${loaded.manifest.name} started`,
-						metadata: { sessionId: _sessionId },
-					});
+					// Emit once even across fallback attempts.
+					if (!sessionStartEmitted) {
+						sessionStartEmitted = true;
+						pushMsg({
+							type: "system",
+							subtype: "session_start",
+							content: `Agent ${loaded.manifest.name} started`,
+							metadata: { sessionId: _sessionId },
+						});
+					}
 					break;
 
 				case "message_update": {
@@ -329,6 +349,7 @@ export function query(options: QueryOptions): Query {
 						_llmCallStart = Date.now();
 					}
 					if (e.type === "text_delta") {
+						attempt.producedOutput = true;
 						accText += e.delta;
 						pushMsg({
 							type: "delta",
@@ -336,6 +357,7 @@ export function query(options: QueryOptions): Query {
 							content: e.delta,
 						});
 					} else if (e.type === "thinking_delta") {
+						attempt.producedOutput = true;
 						accThinking += e.delta;
 						pushMsg({
 							type: "delta",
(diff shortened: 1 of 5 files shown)


shreyas-lyzr (approved):
Clean, well-scoped fix for a genuine gap — the fallback list was always dead config, and this wires it up correctly.

A few things I looked at closely:

**resolveModel() extraction (loader.ts)**: The refactor is faithful — the new function is identical logic to what was inline before, just reusable. The side-effecting mutation of `process.env.OPENAI_API_KEY` and `model.provider` is a pre-existing pattern carried forward, not introduced here, so no regression there.

**Retry classifier (model-fallback.ts)**: The `RETRYABLE_PATTERNS` list is conservative and correct. The choice to retry on empty/null errors (give the next model a chance on opaque failures) is reasonable and documented. One minor note: `/\b500\b/` and `/\b502\b/` etc. correctly use word boundaries so they won't spuriously match numbers embedded in longer strings (e.g. "1500 tokens") — verified.

**Transcript rollback (sdk.ts, line 560–573)**: The shallow array copy (`[...agent.state.messages]`) is appropriate here because message arrays are append-only in pi-agent-core — new turns are pushed, not mutated in-place. A rollback to the snapshot correctly drops the failed user+assistant turn before the retry.

**session_start deduplication**: The `sessionStartEmitted` guard correctly prevents duplicate session-start events across fallback retries. Good.

**finalizeRun() guard**: The `runFinalized` flag prevents double-emission of session_end when the outer `channel.finish()` call at the bottom of the async block also runs. Correct.

**Tests**: The classifier tests cover billing, availability, auth, empty, and request-shaped errors. The manifest resolution tests cover dedup and empty-string skipping. Good coverage of the new logic paths.

Overall the implementation does exactly what it says, is well-tested, and doesn't break existing callers. Approved.

src/sdk.ts
```diff
@@ -474,17 +501,112 @@ export function query(options: QueryOptions): Query {
...
+		// model fails with a retryable provider error before producing any output.
+		const runPrompt = async (promptText: string): Promise<void> => {
+			for (let i = activeModelIndex; i < candidateModels.length; i++) {
+				attempt = { producedOutput: false, error: null };
+				if (i !== activeModelIndex) {
+					agent.state.model = candidateModels[i];
+					activeModelIndex = i;
+				}
+				// Snapshot transcript so a failed attempt can be rolled back cleanly.
+				const snapshot = [...agent.state.messages];
```
shreyas-lyzr, line 562: The snapshot is taken before agent.prompt() is called, so it captures the pre-attempt state. This is correct only if pi-agent-core's prompt() appends the user message synchronously before the first await. If that ordering is not guaranteed by the library's contract, the snapshot could race. Worth a comment confirming the assumption, or a defensive length-check after the failure.

src/sdk.ts
```diff
@@ -474,17 +501,112 @@ export function query(options: QueryOptions): Query {
...
+		// One persistent agent for the whole run so multi-turn conversation state
+		// survives across prompts. Fallback swaps the model on this same agent
+		// (and restores the pre-attempt transcript) rather than rebuilding it.
+		const agent = buildAgent(candidateModels[0]);
+		let activeModelIndex = 0;
+
+		// Run one prompt, falling back through candidate models when the current
+		// model fails with a retryable provider error before producing any output.
+		const runPrompt = async (promptText: string): Promise<void> => {
+			for (let i = activeModelIndex; i < candidateModels.length; i++) {
```
shreyas-lyzr, line 555: activeModelIndex is not reset between prompts in a multi-turn session. After a fallback on turn 1, turns 2..N all start from the fallback model, never retrying the primary even if it has recovered. If this is intentional (sticky fallback), add a comment explaining the choice.

shreyas-lyzr (changes requested):
Three issues worth addressing before merge — one is a correctness gap in the fallback retry logic, one is a state-management risk in multi-turn sessions, and one is a test coverage gap on the actual retry path.

**1. Correctness: snapshot rollback does not include the failed user message turn**

In runPrompt, the snapshot is taken after agent.prompt(promptText) is called, so the snapshot captures the state *before* the user turn was appended internally by the library. After a failed attempt the code sets agent.state.messages = snapshot, which should drop the partially-appended turn. However, the library's prompt() call may append the user message to agent.state.messages *before* awaiting the LLM — in which case snapshot = [...agent.state.messages] taken before agent.prompt() captures the clean state and the rollback is correct. This ordering is library-internal and not guaranteed by the API surface. If pi-agent-core appends the user message asynchronously (e.g. inside the promise after the first await), the snapshot could include it, and after rollback the retry re-sends the same user message — leaving a duplicate in the transcript. This needs a confirmed read of the pi-agent-core prompt() internals or a defensive snapshot taken before *and* the messages length checked after failure.

**2. State risk: activeModelIndex persists across multi-turn prompts after a fallback**

Once a fallback fires and activeModelIndex advances, all subsequent turns in a multi-turn session run on the fallback model — which is the intended behaviour. But the fallback model receives the conversation history from the failed attempt only after the rollback drops the errored user+assistant pair. If the same prompt is retried on the fallback and succeeds, the transcript is correct. However, if a *later* prompt (turn N+1 in the outer loop) triggers another retryable error, the loop starts from activeModelIndex (the current fallback), not 0. This means the primary model is never tried again after a single fallback event, even if it has recovered. This is a design choice, but it is not documented and could surprise users who expect the primary to be re-tried on each turn. If the intent is 'sticky fallback', document it explicitly.

**3. Test gap: the end-to-end retry path (canRetry=true → next model → success) has no test**

The tests in model-fallback.test.ts cover isRetryableProviderError and loadAgent fallback resolution, but there is no test that drives the runPrompt retry loop itself — i.e. a test where model[0] fails with a retryable error and model[1] succeeds. The retry logic in sdk.ts is the riskiest part of this PR (snapshot, rollback, model swap, agent.state mutation) and it is untested. A unit or integration test that mocks agent.prompt() to fail once then succeed on the second call would catch regressions here.

Minor: the fallback system message content string uses template literal interpolation for provider/model but falls back to empty string when those fields are undefined. A provider that omits these would produce 'Model : failed (...)' which is confusing. Defaulting to 'unknown' (as done in the terminal-failure path) would be consistent.

Security pass: no new dependencies introduce known CVEs, no hard-coded secrets, no injection sinks. isRetryableProviderError operates on strings it receives, not on user-controlled input that reaches a sink. Clean.

PR #71, opened 2026-07-06
test: add unit coverage for cost-tracker, compact, knowledge, tool-utils

stealthwhizz:
## Summary

Several core modules shipped with **no unit tests**. This PR adds deterministic, offline `node:test` suites for the pure logic in four of them — **+26 test cases**, all placed under `test/` so `npm test` actually runs them.

## Coverage before / after

Existing tests covered `sdk` (tool helper, schema builder, programmatic hooks, query error paths) and `telemetry`. These modules had none:

| Module | New cases | What's covered |
|---|---|---|
| `cost-tracker.ts` | 6 | Session + per-model token/cost aggregation, derived `totalTokens` fallback, defensive `get()` copy, `reset()` |
| `compact.ts` | 10 | `estimateTokens` heuristic, per-message-type accounting (incl. tool_use overhead), 75% compaction threshold, tool-result head/tail truncation, transcript rendering (drops deltas/system noise), compact-prompt building |
| `knowledge.ts` | 6 | Index loading: `always_load` preload + trim, available listing, missing-file skip, malformed index → empty; prompt formatting (inline docs + read hint) |
| `tool-utils.ts` | 4 | `GCToolDefinition → AgentTool` mapping, string vs `{text, details}` handler results, abort-signal forwarding |

## Notes

- All suites are pure/offline — no API keys, no network, temp dirs cleaned up.
- Full suite: **53 passing** (27 existing + 26 new).

### Follow-up worth flagging (not in this PR)
The `npm test` script globs only `test/*.test.ts`, so the existing suites under `src/**/__tests__/` (memory, sandbox-memory, skill-learner, task-tracker) and `src/__tests__/telemetry.test.ts` are **never run by CI**. Wiring those into the test script (or moving them under `test/`) would be a good separate cleanup.


shreyas-lyzr (approved):
26 new test cases across four previously-uncovered modules. All look correct.

A few things I checked explicitly:

- estimateTokens arithmetic (ceil(5/4)=2), estimateMessageTokens summation (1+2+51+1=55), and the 75% compaction threshold (800/1000=0.8) all hold.
- totalTokens fallback (input+output when field absent) is tested correctly.
- The defensive-copy test for CostTracker.get() properly verifies internal state survives external mutation of the snapshot.
- knowledge tests use mkdtemp + rm recursive cleanup — no leaked temp dirs.
- before() blocks are correctly placed at module scope, not inside describe(), which is required for node:test.
- All suites are offline — no network, no API keys, no side effects.

The follow-up note in the PR description about test/__tests__ not being wired into the npm test glob is a real gap worth tracking separately.

shreyas-lyzr (approved):
Good addition. Reviewed the four new test suites against their source implementations and ran the full suite locally after building.

Verified:
- cost-tracker: all six cases map cleanly to the implementation. The defensive-copy test correctly validates that key deletion on a snapshot doesn't corrupt internal state (shallow copy of modelUsage object is sufficient for that guarantee).
- compact: token arithmetic checks out — ceil(5/4)=2, tool_use overhead of 50 added, needsCompaction threshold at 0.75 all match the source. truncateToolResults head/tail splitting and messagesToText filtering (drops delta/system) are correctly exercised.
- knowledge: temp-dir setup/teardown is clean, covers all four index loading paths (missing, always_load, missing file skip, malformed YAML), and the formatKnowledgeForPrompt assertion matches the exact tag/attribute ordering the source produces.
- tool-utils: covers the two handler result shapes (string vs object with details) and abort signal forwarding, which are the only real branches in toAgentTool.
- mcp.test.ts fixes: the four changed assertions update from the old string return shape to the current content[0].text shape — correct.

Full suite: 65 pass, 0 fail.

Security pass: no new dependencies added, no secrets or credentials in the diff, no injection surface (pure offline unit tests with temp dirs).

One thing worth a follow-up (noted in the PR description): the existing suites under src/**/__tests__/ and src/__tests__/telemetry.test.ts are not picked up by the test glob and never run in CI. Worth wiring those in separately.

PR #72, opened 2026-07-06
chore(test): run all test suites, not just test/*.test.ts

stealthwhizz:
## Problem

The test script globbed only `test/*.test.ts`:

```json
"test": "node --test test/*.test.ts --experimental-strip-types"
```

So the suites under `src/**/__tests__/` (memory, sandbox-memory, skill-learner, task-tracker) and `src/__tests__/telemetry.test.ts` were **never executed by `npm test` or CI** — silently non-running tests.

## What I checked first

Ran those orphaned files directly. They're currently `it.todo(...)` **placeholder stubs** with no bodies, so they report as TODO markers, not failures. That means flipping the glob is safe today (CI stays green) and, more importantly, any *real* tests added under those paths from now on will actually run instead of being silently ignored.

## Fix

```json
"test": "node --test --experimental-strip-types \"test/**/*.test.ts\" \"src/**/__tests__/*.test.ts\""
```

Recursive globs covering both locations.

## Result

`npm test` now discovers every suite:

```
ℹ tests 45
ℹ pass 27
ℹ fail 0
ℹ todo 18
```

27 real tests pass, 18 TODO stubs surface as TODO (not failures), 0 failures.

Follow-up flagged from #71; kept as its own one-line change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "build": "tsc",
     "dev": "tsc --watch",
     "start": "node dist/index.js",
-    "test": "node --test test/*.test.ts --experimental-strip-types"
+    "test": "node --test --experimental-strip-types \"test/**/*.test.ts\" \"src/**/__tests__/*.test.ts\""
   },
   "engines": {
     "node": ">=20"


shreyas-lyzr (approved):
The fix is correct. The previous glob silently dropped all tests under src/**/__tests__/ and any nested test/ paths. Switching to two explicit recursive globs is the right approach, and the author already verified CI stays green since the newly discovered suites are todo stubs. Nothing blocking.

shreyas-lyzr (approved):
Correct and minimal fix. The old glob (test/*.test.ts) was silently skipping all suites under src/**/__tests__/ and the nested test/**/*.test.ts subdirectory. The new command covers both locations with explicit recursive globs. Node's --test runner accepts multiple glob arguments and handles duplicates gracefully, so there is no risk of double-counting.

The PR description correctly notes that the newly-discovered suites are all it.todo() stubs, so CI stays green immediately after merge. This is a safe landing point.

One small observation: the glob order matters if any test suite has a --test-reporter or ordering dependency, but given these are Node's built-in test runner suites with no custom reporter config, the order is inconsequential.

Security pass: package.json script change only, no new dependencies, no credential exposure. Clean.

PR #73, opened 2026-07-07
fix: compute costUsd from model rates when the provider reports none (#67)

### LINKED ISSUE #67: bug: costs() always returns 0 — token usage tracked but costUsd never calculated correctly
## Summary

Token counts are read correctly from the API response, but `costs()` always returns `costUsd: 0` regardless of provider or model. Reported and investigated by @stealthwhizz.

---

## Root cause analysis

There are **three distinct bugs** across the stack, each contributing to the same symptom.

### Bug 1 — `EMPTY_USAGE` zeros out costs on the error path (`agent.ts:333`)

`EMPTY_USAGE` is a module-level constant applied exclusively inside `handleRunFailure()`:

```ts
usage: EMPTY_USAGE,   // line 333 of agent.js
```

On **successful** runs, `msg.usage` is fully populated. On **error/aborted** runs, usage is silently zeroed — meaning `costs()` returns 0 even when tokens were consumed before the error. This is arguably intentional but is completely undocumented, so callers have no way to distinguish "nothing was billed" from "billing data was lost".

### Bug 2 — `calculateCost()` in pi-ai silently emits 0 for unknown model IDs

`calculateCost()` depends on `model.cost.input` / `model.cost.output` being present in the model manifest. For custom, unrecognized, or manifest-less model IDs, these fields are zero or undefined, so `cost.total = 0`. `sdk.ts` faithfully reads that 0:

```ts
costUsd: msg.usage.cost?.total ?? 0,
```

There is no warning, no log, no throw — `costs()` silently returns 0 with real `inputTokens` / `outputTokens`. This is the primary symptom most users will hit.

### Bug 3 — No cost calculation for OpenAI completions streaming

The `openai-completions` provider reads token counts from `stream_options.include_usage` but never calls `calculateCost()` or equivalent. Only `openai-responses` has the `applyServiceTierPricing` flow. So `costUsd` is always 0 for the completions API path regardless of model.

---

## Affected files

| File | Issue |
|---|---|
| `agent.ts` (pi-agent-core) | `EMPTY_USAGE` on error path — undocumented, misleading |
| `sdk.ts` (gitagent) | Should warn/throw when `inputTokens > 0` but `costUsd == 0` |
| `openai-completions` provider (pi-ai) | Missing `calculateCost()` call after usage parsing |

---

## Proposed fixes

1. **`sdk.ts`** — Add a guard: if `inputTokens > 0 && costUsd === 0`, emit a warning (or surface it in `costs()` metadata) so callers know cost data is unavailable rather than literally zero.

2. **pi-ai / `calculateCost()`** — Return `null` (not `0`) when pricing data is missing for a model ID, so callers can distinguish "calculated zero cost" from "no pricing data available". Log a warning with the unrecognized model ID.

3. **`openai-completions` provider** — Call `calculateCost()` (or equivalent) after parsing usage tokens, consistent with the `openai-responses` path.

4. **`handleRunFailure()`** — Document that `EMPTY_USAGE` is intentional on the error path, or accumulate partial usage before the failure and report it.

---

## Reproduction

Any run using a custom model ID, or using the `openai-completions` API path — check that `costs().totalCostUsd` is non-zero after a successful multi-turn session. It will be 0.

---

Reported by @stealthwhizz.

stealthwhizz:
## Problem

`costs()` reports `costUsd: 0` even when tokens were consumed and billed. `sdk.ts` mapped cost straight from the provider:

```ts
costUsd: msg.usage.cost?.total ?? 0,
```

But `msg.usage.cost.total` is only populated when pi-ai's `calculateCost` runs, which:
- **doesn't run on the `openai-completions` path** (Bug 3 in #67), and
- **can't produce a real number for custom/unregistered models** that carry no price table (Bug 2).

So the 0 was read verbatim, and callers couldn't tell "this was free" from "we don't know the cost."

## Root-cause split

| Bug | Where | Fixable in gitagent? |
|---|---|---|
| `calculateCost` emits 0 for unknown model IDs | pi-ai | ❌ upstream |
| `openai-completions` never calls `calculateCost` | pi-ai | ❌ upstream |
| `EMPTY_USAGE` zeros usage on the error path | pi-agent-core | ❌ upstream |
| `sdk.ts` reports that 0 with no fallback / no signal | **gitagent** | ✅ this PR |

## Fix (gitagent side)

- **`computeCostUsd(usage, rates)`** in `cost-tracker.ts` — the same per-million-token formula pi-ai uses (`(rate / 1e6) * tokens`). Returns `null` when tokens were used but the model has no pricing, so "unknown" is distinguishable from `0`.
- **`sdk.ts`** — prefer the provider's `cost.total` when it's present and non-zero; otherwise recompute from `loaded.model.cost` (known models carry real rates, e.g. gpt-4o = \$2.5/\$10 per 1M). Pass a `costResolved` flag to the tracker.
- **`SessionCosts.costDataAvailable`** — flips to `false` when any request consumed tokens that couldn't be priced, so `costs()` consumers can surface "cost unavailable" instead of a misleading \$0.
- Exported `computeCostUsd` + the new types.

## Result

- Known models that hit the `openai-completions` path (or otherwise arrive with no provider cost) now get a correctly computed `costUsd` from their own rate table.
- Custom/zero-rate models report `costDataAvailable: false` instead of silently claiming \$0.
- Token counts are untouched; this only changes how cost is derived and reported.

## Tests

`test/cost-compute.test.ts` — pricing math (input/output/cache), `null` for unpriceable-with-tokens, `0` for no-tokens-no-rates, partial-rate models; plus `costDataAvailable` default/flip/reset. Full suite green.

## Upstream follow-ups (out of scope here)
Worth filing against pi-ai / pi-agent-core: have `calculateCost` return `null` (not 0) for missing pricing, call it on the `openai-completions` path, and document/​accumulate `EMPTY_USAGE` on the error path.

Closes #67

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -22,7 +22,7 @@ import type {
 	QueryOptions,
 	SandboxOptions,
 } from "./sdk-types.js";
-import { CostTracker } from "./cost-tracker.js";
+import { CostTracker, computeCostUsd } from "./cost-tracker.js";
 import { context as otelContext } from "@opentelemetry/api";
 import {
 	wrapToolWithOtel,
@@ -370,6 +370,35 @@ export function query(options: QueryOptions): Query {
 
 					const { text, thinking } = extractContent(msg);
 
+					// Resolve cost. Providers that call pi-ai's calculateCost supply
+					// msg.usage.cost.total; the openai-completions path and custom
+					// models do not, leaving it 0 even when tokens were billed
+					// (issue #67). When it's missing, recompute from the model's own
+					// per-million-token rates. costResolved is false only when tokens
+					// were used but no pricing exists — so callers can distinguish
+					// "unknown" from a real $0.
+					let costUsd = 0;
+					let costResolved = true;
+					if (msg.usage) {
+						const tokens = {
+							inputTokens: msg.usage.input ?? 0,
+							outputTokens: msg.usage.output ?? 0,
+							cacheReadTokens: msg.usage.cacheRead ?? 0,
+							cacheWriteTokens: msg.usage.cacheWrite ?? 0,
+						};
+						const providerCost = msg.usage.cost?.total ?? 0;
+						if (providerCost > 0) {
+							costUsd = providerCost;
+						} else {
+							const computed = computeCostUsd(tokens, loaded.model.cost);
+							if (computed === null) {
+								costResolved = false; // tokens used, no pricing → unknown
+							} else {
+								costUsd = computed;
+							}
+						}
+					}
+
 					const assistantMsg: GCAssistantMessage = {
 						type: "assistant",
 						content: text || accText,
@@ -384,7 +413,7 @@ export function query(options: QueryOptions): Query {
 							cacheReadTokens: msg.usage.cacheRead ?? 0,
 							cacheWriteTokens: msg.usage.cacheWrite ?? 0,
 							totalTokens: msg.usage.totalTokens ?? 0,
-							costUsd: msg.usage.cost?.total ?? 0,
+							costUsd,
 						} : undefined,
 					};
 					pushMsg(assistantMsg);
@@ -394,6 +423,7 @@ export function query(options: QueryOptions): Query {
 						costTracker.add(
 							`${assistantMsg.provider}:${assistantMsg.model}`,
 							assistantMsg.usage,
+							costResolved,
 						);
 						_totalCostUsd += assistantMsg.usage.costUsd ?? 0;
 					}
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -387,10 +387,14 @@ export function query(options: QueryOptions): Query {
 							cacheWriteTokens: msg.usage.cacheWrite ?? 0,
 						};
 						const providerCost = msg.usage.cost?.total ?? 0;
+						const rates = loaded.model.cost;
 						if (providerCost > 0) {
 							costUsd = providerCost;
+						} else if (!rates) {
+							// Model object carries no cost table at all — can't price it.
+							costResolved = false;
 						} else {
-							const computed = computeCostUsd(tokens, loaded.model.cost);
+							const computed = computeCostUsd(tokens, rates);
 							if (computed === null) {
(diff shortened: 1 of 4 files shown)


src/sdk.ts
```diff
@@ -370,6 +370,35 @@ export function query(options: QueryOptions): Query {
...
+							inputTokens: msg.usage.input ?? 0,
+							outputTokens: msg.usage.output ?? 0,
+							cacheReadTokens: msg.usage.cacheRead ?? 0,
+							cacheWriteTokens: msg.usage.cacheWrite ?? 0,
+						};
+						const providerCost = msg.usage.cost?.total ?? 0;
+						if (providerCost > 0) {
+							costUsd = providerCost;
+						} else {
+							const computed = computeCostUsd(tokens, loaded.model.cost);
```
shreyas-lyzr: loaded.model.cost is always set today (createCustomModel hardcodes it, getModel from pi-ai includes it), so this won't throw in practice. Worth adding a null-fallback anyway so a future model source that omits the field degrades gracefully rather than crashing:

```suggestion
							const computed = computeCostUsd(tokens, loaded.model.cost ?? { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 });
```

shreyas-lyzr (approved):
Clean fix for the misleading $0 cost reporting. The three-way outcome — provider cost wins when non-zero, fallback computation from model rates, and null/costDataAvailable=false for genuinely unresolvable cases — is the right design. The logic in computeCostUsd correctly mirrors pi-ai's formula, and the test coverage hits all the meaningful branches including the partial-rate and zero-token edges.

One suggestion below: null-guard loaded.model.cost in sdk.ts. It's safe today because both createCustomModel and getModel always produce a cost object, but a future model source that omits it would throw at runtime rather than fail gracefully.

shreyas-lyzr (approved):
Re-approving to cover the new commit. The null-guard on loaded.model.cost is exactly the right fix — plugin-provided models without a cost table now cleanly set costDataAvailable: false instead of throwing. Both commits together are solid; ready to merge.

PR #75, opened 2026-07-07
fix: Windows compatibility for hooks and cli tool

### LINKED ISSUE #74: Hook path-safety check is Windows-incompatible, causing pre_tool_use hooks to silently no-op
## Summary
`pre_tool_use` hooks (e.g. a safety-check script meant to block destructive commands like `rm -rf`) silently fail to run on Windows, because the base-directory containment check in `hooks.ts` is hardcoded to POSIX path separators.

## Root cause
`src/hooks.ts` (around line 69):

```js
if (!resolvedScript.startsWith(allowedBase + "/") && resolvedScript !== allowedBase) {
    reject(new Error(`Hook "${hook.script}" escapes its base directory`));
}
```
allowedBase + "/" hardcodes a forward slash. On Windows, path.resolve() returns backslash-separated paths (e.g. C:\Users\...\hooks\safety-check.sh), so this check always fails for a legitimately-nested hook script — it's a false positive, not a real path-traversal case.

Compounding this: runHooks() catches any hook execution error and defaults to { action: "allow" } (fail-open), so the false "escapes its base directory" error is silently swallowed and the tool call proceeds as if no hook existed at all.

## Impact
On Windows, pre_tool_use hooks never actually execute their logic. Any safety/validation hook (e.g. blocking rm -rf) is effectively disabled, with only a console warning (Hook error: ... escapes its base directory) as a trace — the destructive command still runs.

Repro

import path from "path";
const { join, resolve } = path.win32;
const allowedBase = resolve("C:\\Users\\test\\agent");
const resolvedScript = resolve(join("C:\\Users\\test\\agent", "hooks", "safety-check.sh"));
console.log(resolvedScript.startsWith(allowedBase + "/")); // false — bug reproduced without needing Windows
## Fix
Replace the manual string check with path.relative(), which correctly handles OS-specific separators:

const rel = relative(allowedBase, resolvedScript);
const escapesBase = rel !== "" && (rel.startsWith("..") || isAbsolute(rel));
Verified this resolves correctly on both path.win32 and path.posix, while still correctly catching genuine traversal attempts (e.g. ../../evil.sh).

Separately worth addressing: runHooks()'s fail-open behavior on any hook error — a safety-oriented hook failing to execute for any reason (not just this bug) currently defaults to allowing the action rather than blocking it.

Nivesh353:
## Summary
Fixes two Windows-specific bugs that caused `pre_tool_use` hooks (e.g. a
safety-check blocking `rm -rf`) to silently fail on Windows, and the `cli`
tool to fail entirely outside Git Bash/WSL.

- **Hook path-containment check was hardcoded to POSIX separators.** The
  check testing whether a hook script stays inside its base directory did
  `resolvedScript.startsWith(allowedBase + "/")` — hardcoding `/`. On
  Windows, `path.resolve()` returns backslash-separated paths, so this
  check always failed, even for correctly-nested scripts. Combined with
  `runHooks()`'s fail-open error handling, this silently disabled every
  hook on Windows with no clear indication why. Replaced with
  `path.relative()`, which correctly handles OS-specific separators on
  Windows, macOS, and Linux.

- **`cli` tool hardcoded `spawn("sh", ["-c", command])`.** `sh` isn't
  available on native Windows (PowerShell/cmd) without Git Bash or WSL
  installed, so every `cli` tool call failed with `spawn sh ENOENT` there.
  Switched to `spawn(command, { shell: true })`, so Node picks the correct
  native shell per OS (`cmd.exe` on Windows, `/bin/sh` on macOS/Linux) —
  no behavior change on macOS/Linux, since that already resolves to
  `/bin/sh`.

- Also added a clearer error message when a hook script genuinely can't
  run because no POSIX shell is available at all (e.g. native PowerShell
  without Git Bash/WSL), instead of a raw `ENOENT`.

## Test plan
- [x] Reproduced both bugs locally by forcing Windows path semantics
      (`path.win32`) and a stripped `PATH` without requiring an actual
      Windows machine
- [x] Verified `path.relative()` fix passes for correctly-nested Windows
      paths and still correctly rejects genuine traversal attempts
- [x] Confirmed on an actual Windows machine (Git Bash): `cli` tool and
      hook execution both work correctly after the fix
- [x] Confirmed on macOS: no behavior change for either the `cli` tool or
      hook execution (regression check)

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,6 +1,6 @@
 import { spawn } from "child_process";
 import { readFile } from "fs/promises";
-import { join, resolve, relative, isAbsolute } from "path";
+import { join, resolve, relative, isAbsolute, sep } from "path";
 import yaml from "js-yaml";
 import type { AgentTool } from "@mariozechner/pi-agent-core";
 
@@ -67,7 +67,7 @@ async function executeHook(
 		const resolvedScript = resolve(scriptPath);
 		const allowedBase = resolve(baseDir);
 		const rel = relative(allowedBase, resolvedScript);
-		const escapesBase = rel !== "" && (rel.startsWith("..") || isAbsolute(rel));
+		const escapesBase = rel !== "" && (rel === ".." || rel.startsWith(".." + sep) || isAbsolute(rel));
 		if (escapesBase) {
 			reject(new Error(`Hook "${hook.script}" escapes its base directory`));
 			return;
@@ -99,7 +99,7 @@ async function executeHook(
 
 		child.on("error", (err: any) => {
 			clearTimeout(timeout);
-			if (err.code === "ENOENT" && err.syscall === "spawn sh") {
+			if (err.code === "ENOENT" && err.path === "sh") {
 				reject(new Error(
 					`Hook "${hook.script}" could not run: no POSIX shell ("sh") found on this system. ` +
 					`Hook scripts require a POSIX shell — on Windows, run gitagent from Git Bash or WSL, ` +
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -97,8 +97,16 @@ async function executeHook(
 			reject(new Error(`Hook "${hook.script}" timed out after 10s`));
 		}, 10_000);
 
-		child.on("error", (err) => {
+		child.on("error", (err: any) => {
 			clearTimeout(timeout);
+			if (err.code === "ENOENT" && err.syscall === "spawn sh") {
+				reject(new Error(
+					`Hook "${hook.script}" could not run: no POSIX shell ("sh") found on this system. ` +
+					`Hook scripts require a POSIX shell — on Windows, run gitagent from Git Bash or WSL, ` +
+					`or ensure "sh" is on PATH.`,
+				));
+				return;
+			}
 			reject(new Error(`Hook "${hook.script}" failed to start: ${err.message}`));
 		});
 
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,6 +1,6 @@
 import { spawn } from "child_process";
 import { readFile } from "fs/promises";
-import { join, resolve } from "path";
+import { join, resolve, relative, isAbsolute } from "path";
 import yaml from "js-yaml";
 import type { AgentTool } from "@mariozechner/pi-agent-core";
 
@@ -66,7 +66,9 @@ async function executeHook(
 		// Path traversal guard: ensure script doesn't escape its base directory
 		const resolvedScript = resolve(scriptPath);
 		const allowedBase = resolve(baseDir);
-		if (!resolvedScript.startsWith(allowedBase + "/") && resolvedScript !== allowedBase) {
+		const rel = relative(allowedBase, resolvedScript);
+		const escapesBase = rel !== "" && (rel.startsWith("..") || isAbsolute(rel));
+		if (escapesBase) {
 			reject(new Error(`Hook "${hook.script}" escapes its base directory`));
 			return;
 		}
--- a/src/tools/cli.ts
+++ b/src/tools/cli.ts
@@ -24,10 +24,11 @@ export function createCliTool(cwd: string, defaultTimeout?: number): AgentTool<t
 					return;
 				}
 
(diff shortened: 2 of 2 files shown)


src/hooks.ts
```diff
@@ -95,8 +97,16 @@ async function executeHook(
...
 
-		child.on("error", (err) => {
+		child.on("error", (err: any) => {
 			clearTimeout(timeout);
+			if (err.code === "ENOENT" && err.syscall === "spawn sh") {
+				reject(new Error(
+					`Hook "${hook.script}" could not run: no POSIX shell ("sh") found on this system. ` +
+					`Hook scripts require a POSIX shell — on Windows, run gitagent from Git Bash or WSL, ` +
+					`or ensure "sh" is on PATH.`,
+				));
```
shreyas-lyzr, line 107: The condition `err.syscall === "spawn sh"` is likely wrong. Node sets `err.syscall` to `"spawn"` on ENOENT spawn errors, not `"spawn sh"`. You can verify with a quick test: `spawn("sh", ["-c", "echo hi"])` on a machine without `sh` — the syscall field will be `"spawn"`, not `"spawn sh"`. If this condition never matches, the improved error message is dead code and users still see the raw ENOENT. Fix: check `err.code === "ENOENT"` alone, or also check `err.path === "sh"` which Node does populate reliably.

src/hooks.ts
```diff
@@ -66,7 +66,9 @@ async function executeHook(
 		// Path traversal guard: ensure script doesn't escape its base directory
 		const resolvedScript = resolve(scriptPath);
 		const allowedBase = resolve(baseDir);
-		if (!resolvedScript.startsWith(allowedBase + "/") && resolvedScript !== allowedBase) {
+		const rel = relative(allowedBase, resolvedScript);
```
shreyas-lyzr, line 69: `rel.startsWith("..")` will reject paths like `..legit-dir/script.sh` (a directory whose name starts with `..foo`) as escaping — a false positive. Since this is a security guard, prefer the more precise check: `rel === ".." || rel.startsWith(".." + path.sep)`. The import for `sep` would need to be added. Low-likelihood false positive in practice, but worth tightening given the security context.

src/tools/cli.ts
```diff
@@ -24,10 +24,11 @@ export function createCliTool(cwd: string, defaultTimeout?: number): AgentTool<t
 					return;
 				}
 
-				const child = spawn("sh", ["-c", command], {
+				const child = spawn(command, {
```
shreyas-lyzr, line 30: Worth a comment noting that `shell: true` on Windows routes through `cmd.exe`, which has different quoting/escaping rules than `sh`. Characters like `&`, `|`, `^`, `>`, and embedded double-quotes behave differently. This is probably acceptable for a tool that already runs arbitrary commands, but document it so future maintainers know Windows behavior may diverge from Linux/macOS in edge cases.

shreyas-lyzr (changes requested):
Three issues worth addressing before merge — one is a likely dead code path (the improved error message may never fire), one is a precision issue in a security check, and one is a behavior note for the spawn change.

src/hooks.ts
Nivesh353: Fixed — now checks err.path === 'sh' alongside err.code === 'ENOENT', matching what Node actually sets. Verified the friendly message now fires correctly.

src/hooks.ts
Nivesh353: Fixed — now uses rel === '..' || rel.startsWith('..' + sep) instead of a bare startsWith('..'). Verified: ..foo/script.sh now allowed, ../evil.sh still blocked.

src/tools/cli.ts
Nivesh353: Added a comment documenting that cmd.exe's quoting/escaping rules differ from sh on Windows, so this is clear for future maintainers."

shreyas-lyzr (approved):
All three issues from the previous review are addressed correctly. The syscall check is now using err.path === "sh" as suggested, the traversal guard uses path.sep for precision and handles the absolute-path case too, and the spawn change is documented with an inline comment about cmd.exe quoting differences. Good to merge.

PR #77, opened 2026-07-13
fix(security): resolve Aikido findings — command injection, path traversal, prototype pollution

Nivesh353:
## Summary

Addresses the security findings from the Aikido scan by hardening how
untrusted input (tool args, plugin manifests, config values, branch/skill
names) is handled across the codebase. No user-facing feature changes.

## What changed

**Command injection (CWE-78)**
Replaced shell-interpolated `execSync("git …")` with
`execFileSync("git", [args])` so names/paths can't break out into shell
commands.
- `session.ts` — all git ops (clone/fetch/checkout/commit/push)
- `loader.ts` — git clone for `extends`/dependencies
- `tools/capture-photo.ts`, `tools/memory.ts`, `tools/skill-learner.ts` — git add/commit

**Path traversal (CWE-22)**
Confine file access to the intended base directory.
- `knowledge.ts` — skip entries whose path escapes the knowledge dir
- `tool-loader.ts` — reject tool scripts outside `tools/`; add runtime
  allowlist (`sh/bash/node/python3/python`)
- `plugins.ts` — `resolveWithinDir()` guards plugin `entry`/`prompt` paths
- `tools/memory.ts` — memory-layer path must stay within the agent dir
- `chat-history.ts` — `sanitizeBranch()` now strips `\` and `..` (not just
  `/`); exported and reused by `context.ts`

**Input validation**
Require safe kebab-case/identifier names before they're used to build
filesystem paths: `schedules` (id), `workflows` (name), `plugin-cli` +
`plugins` (plugin name), `skill-learner` (skill_name), `task-tracker`
(skill_used), `config` (env), `loader` (dep name).

**Prototype pollution (CWE-1321)**
`deepMerge()` ignores `__proto__` / `constructor` / `prototype` keys in
`config.ts` and `loader.ts`.

**Dependencies / CI hardening**
- `package.json` — `overrides` for `basic-ftp` and `uuid` (patch vulnerable
  transitive deps)
- `publish.yml` — `persist-credentials: false` on checkout; pin Node to `22.14.0`

## Testing
- `tsc` compiles clean
- Existing unit tests pass (27/27)
- Security guards verified against the real shipped tool code (path
  traversal, kebab-case validation, branch sanitizing) — all blocking/
  allowing as expected
- Confirmed the `sanitizeBranch` change does **not** affect the voice web
  UI (branch listing reads real git branches; filenames are unchanged for
  all valid branch names)


src/chat-history.ts
```diff
@@ -6,8 +6,10 @@ import { query } from "./sdk.js";
 /** Types we skip — too large or ephemeral */
 const SKIP_TYPES = new Set(["audio_delta", "agent_thinking"]);
 
-function sanitizeBranch(branch: string): string {
-	return branch.replace(/\//g, "__");
+export function sanitizeBranch(branch: string): string {
+	// Strip both slash types (Windows path.join also treats "\" as a separator)
+	// and any leftover ".." so the result can never traverse outside historyDir.
+	return branch.replace(/[\\/]/g, "__").replace(/\.\./g, "__");
```
shreyas-lyzr: Single-pass replace: '...' (three dots) becomes '__.' because the regex consumes the first two dots and leaves the third. The result never re-introduces a '..' segment so this is not directly exploitable, but an allowlist is easier to audit:

```suggestion
	return branch.replace(/[^a-zA-Z0-9_-]/g, "__");
```

Alternatively, validate the final historyPath join result with resolve/relative like the other guards in this PR.

package.json
```diff
@@ -51,6 +51,7 @@
   "dependencies": {
     "@mariozechner/pi-agent-core": "^0.70.2",
     "@mariozechner/pi-ai": "^0.70.2",
+    "@open-gitagent/voice": "^1.0.0",
```
shreyas-lyzr: This dependency is not present in the base branch and is unrelated to the Aikido security fixes. Its transitive graph includes libsignal (GPL-3.0), baileys, whatsapp-rust-bridge, and sharp. GPL-3.0 in a production dependency is a license concern for a library package. This looks like a rebase artifact — please revert or split into its own PR with an explicit license review.

src/plugins.ts
```diff
@@ -134,7 +144,10 @@ export async function installPlugin(
 	await mkdir(targetDir, { recursive: true });
 
 	// Derive plugin name from source
-	const name = source.split("/").pop()?.replace(/\.git$/, "") || "plugin";
+	const rawName = source.split("/").pop()?.replace(/\.git$/, "") || "plugin";
+	// Strip anything that isn't a safe identifier char — handles backslashes,
+	// "..", etc. that could otherwise survive the forward-slash-only split above.
+	const name = rawName.replace(/[^a-zA-Z0-9_-]/g, "_") || "plugin";
```
shreyas-lyzr: Name sanitized here allows uppercase and underscores (/[^a-zA-Z0-9_-]/g), but KEBAB_RE (/^[a-z0-9]+(-[a-z0-9]+)*$/) used at manifest validation rejects both. A plugin from a URL like https://example.com/My_Plugin.git installs as 'My_Plugin' but may fail KEBAB_RE validation at load time. Normalize to lowercase at install time and use a single naming rule throughout:

```suggestion
	const name = rawName.replace(/[^a-z0-9-]/g, "-").replace(/-+/g, "-").replace(/^-|-$/g, "") || "plugin";
```

shreyas-lyzr (changes requested):
Solid security hardening overall — the execFileSync migrations, deepMerge prototype-pollution guards, and path-traversal checks with resolve/relative are correct and well-applied. A few issues worth addressing before merge.

**Scope creep in package.json**

The PR description says no user-facing feature changes, but package.json adds @open-gitagent/voice as a hard production dependency (not in the base branch) and pulls in a large transitive graph (baileys, whatsapp-rust-bridge, pino, sharp, libsignal/GPL-3.0, etc.). That is unrelated to the Aikido findings and belongs in a separate PR. If this is an accidental rebase artifact it should be reverted. If intentional, it needs its own review including license due diligence on libsignal (GPL-3.0 in a production dependency is a concern for a library codebase).

**knowledge.ts: path traversal guard only covers always_load entries**

The traversal check in the diff protects preloaded entries, but the available entries array (the else branch, line 56 in the file) has no guard. If the read tool later joins knowledgeDir + entry.path without its own validation, a malicious index.yaml could still escape the knowledge directory for on-demand reads. Apply the same resolve/relative check before pushing to available, or confirm that on-demand reads are independently guarded elsewhere.

**sanitizeBranch: non-iterative replacement leaves trailing dot**

The .replace(/\.\./g, '__') is a single pass, so the string '...' (three dots) becomes '__.' — the trailing dot survives. The result does not re-introduce a traversal segment, so this is not directly exploitable as written. However an allowlist replace such as branch.replace(/[^a-zA-Z0-9_-]/g, '__') is simpler to audit, or the final historyPath join result could be validated with resolve/relative like the other guards in this PR.

**Inconsistent name validation between install and load**

installPlugin sanitizes the derived name with /[^a-zA-Z0-9_-]/g (allows uppercase and underscores), while discoverAndLoadPlugins validates with /^[a-zA-Z0-9_-]+$/ (also allows uppercase+underscore), and manifest validation uses KEBAB_RE (/^[a-z0-9]+(-[a-z0-9]+)*$/). A plugin whose repo name contains uppercase letters installs successfully under a name that then fails KEBAB_RE at load time. Normalize to lowercase kebab at install time and enforce a single regex everywhere.

package.json
Nivesh353: this was a rebase/test artifact (an accidental npm install @open-gitagent/voice while testing voice mode), not intended for this PR. Removed the dependency and regenerated

src/plugins.ts
Nivesh353: Applied your suggested change — installPlugin now normalizes the derived directory name to lowercase kebab (My_Plugin → my-plugin)

src/chat-history.ts
Nivesh353: Adopted your suggested allowlist: branch.replace(/[^a-zA-Z0-9_-]/g, "__"). "..." → "______" (no dot survives) and it's easier to audit. Normal branches are unchanged (chat/my-chat → chat__my-chat). Note: single dots are now also replaced (release/1.2.3 → release__1__2__3), which is non-destructive for history and doesn't affect the voice UI (chat branches have no dots). Resolved.

shreyas-lyzr (approved):

1. The @open-gitagent/voice dependency (and its GPL-3.0 transitive chain) is absent from this revision — clean.

2. knowledge.ts path traversal guard now covers the full entry loop before the always_load branch, so on-demand (available) entries are subject to the same resolve/relative check. The original blocking concern is resolved.

3. sanitizeBranch is now an allowlist (/[^a-zA-Z0-9_-]/g → '__'), which is trivially auditable and closes the '...' → '__.' edge case noted earlier.

4. installPlugin normalizes derived names to lowercase kebab-case before writing to disk, so a plugin that installs successfully will always pass KEBAB_RE at load time. The install/load mismatch is gone.

Additional hardening in this revision — execFileSync across all git ops, deepMerge prototype-pollution guards in both config.ts and loader.ts, runtime allowlist in tool-loader.ts, resolveWithinDir() for plugin entry/prompt, input validation on schedule/workflow/skill/dep names — all look correct. The allowlist approach is consistently applied and the guards are placed at the right boundary (before any path join is used for I/O).

No new issues found. Good to merge.

PR #81, opened 2026-07-21
feat: auto-detect API keys and default model for zero-config startup

### LINKED ISSUE #14: Auto-detect API keys and default to latest models for zero-config startup

Nivesh353:
## Problem

gitagent required an explicit model configuration (`model.preferred` in
`agent.yaml`, or `--model` on the CLI) to start — even when a valid
provider API key was already present in the environment. Users had to
manually configure a model before gitagent would run at all.

## Fix

**New file: `src/auto-detect-model.ts`**
Checks known providers in a fixed order and returns a sensible default
model for the first one with usable credentials, using `pi-ai`'s
`getEnvApiKey()` rather than a hardcoded env-var check — so ambient
credential sources are picked up too, not just direct API keys.

Providers checked, in order:
1. Direct API-key providers (explicit, deliberate signal): `anthropic`,
   `openai`, `google`, `xai`, `groq`, `mistral`
2. Ambient/enterprise credential sources (checked after, since these can
   be present incidentally): `amazon-bedrock`, `google-vertex`,
   `azure-openai-responses`, `github-copilot`

All defaults use versionless model aliases (e.g. `claude-sonnet-4-6`,
`gpt-4o`) rather than dated model IDs, so they don't go stale as new
point releases ship.

**`src/loader.ts`**
- Wires `autoDetectModel()` in as the final fallback in model resolution,
  preserving existing priority: `--model` flag > `agent.yaml`
  `model.preferred` > env config override > auto-detected from API key.
- Logs the choice on auto-detect: `Using <model> (auto-detected from
  <provider> credentials)`.
- Updated the "no model configured" error to mention the new option.
- **Related fix:** `manifest.model` is now normalized to `{ fallback: [] }`
  right after parsing `agent.yaml`, and `model.preferred` is now correctly
  typed as optional. Previously, a `model:` block missing entirely (not
  just missing `preferred`) crashed downstream code (`index.ts`, `sdk.ts`)
  with `Cannot read properties of undefined (reading 'constraints')` —
  masked before this change because a missing model always hit the "no
  model configured" error first; auto-detection now gets further and
  exposed it.

`.env` file loading was already implemented separately (`src/index.ts`
loads `.env` from `~/.gitagent/.env` and the agent directory before model
resolution runs), so no changes were needed there.

## Testing

- `npm run build` — compiles cleanly, no type errors.
- `npm test` — all 27 existing tests pass, no regressions.
- Verified live, end-to-end, against a real agent directory:
  - No model configured + a real provider key present → correctly
    auto-detects and logs the choice, proceeds to a real API call.
  - No model configured + no keys at all → correct friendly error,
    no crash.
  - No `model:` block in `agent.yaml` at all → previously crashed with
    `Cannot read properties of undefined (reading 'constraints')`; now
    resolves correctly via auto-detect.

Closes #14

### DIFF
--- a/src/auto-detect-model.ts
+++ b/src/auto-detect-model.ts
@@ -1,0 +1,47 @@
+// Zero-config model auto-detection (issue #14).
+//
+// When no model is configured via --model, agent.yaml, or env config override,
+// check known provider API keys (via pi-ai's getEnvApiKey, so Bedrock / Vertex /
+// Azure / Copilot ambient credentials are picked up too, not just a hardcoded
+// env var list) and fall back to a sensible default model for the first
+// provider that has credentials available.
+
+import { getEnvApiKey } from "@mariozechner/pi-ai";
+
+// Checked in this order; first provider with credentials available wins.
+// Versionless aliases only, so this never goes stale as new point releases ship.
+//
+// Direct API-key providers first (explicit, deliberate signal); ambient/
+// enterprise credential sources (Bedrock, Vertex, Azure, Copilot) checked
+// after, since those can be present incidentally (e.g. a leftover AWS_PROFILE
+// or an active gcloud login unrelated to this agent).
+const DEFAULT_MODEL_BY_PROVIDER: Array<{ provider: string; model: string }> = [
+	{ provider: "anthropic", model: "claude-sonnet-4-6" },
+	{ provider: "openai", model: "gpt-4o" },
+	{ provider: "google", model: "gemini-2.0-flash" },
+	{ provider: "xai", model: "grok-4-fast" },
+	{ provider: "groq", model: "llama-3.3-70b-versatile" },
+	{ provider: "mistral", model: "mistral-large-latest" },
+	{ provider: "amazon-bedrock", model: "anthropic.claude-sonnet-4-6" },
+	{ provider: "google-vertex", model: "gemini-2.0-flash" },
+	{ provider: "azure-openai-responses", model: "gpt-4o" },
+	{ provider: "github-copilot", model: "claude-sonnet-4.6" },
+];
+
+export interface AutoDetectedModel {
+	modelStr: string;
+	provider: string;
+}
+
+/**
+ * Find the first provider with a usable API key/credential in the environment
+ * and return its default model, or undefined if none are configured.
+ */
+export function autoDetectModel(): AutoDetectedModel | undefined {
+	for (const { provider, model } of DEFAULT_MODEL_BY_PROVIDER) {
+		if (getEnvApiKey(provider)) {
+			return { modelStr: `${provider}:${model}`, provider };
+		}
+	}
+	return undefined;
+}
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -34,7 +34,7 @@ export interface AgentManifest {
 	tags?: string[];
 	metadata?: Record<string, string | number | boolean>;
 	model: {
-		preferred: string;
+		preferred?: string;
 		fallback: string[];
 		constraints?: {
 			temperature?: number;
@@ -242,6 +242,9 @@ export async function loadAgent(
 	// Parse agent.yaml
 	const manifestRaw = await readFile(join(agentDir, "agent.yaml"), "utf-8");
 	let manifest = yaml.load(manifestRaw) as AgentManifest;
+	// model: is optional in agent.yaml (auto-detection can fill it in later) —
+	// normalize here so downstream code can always assume manifest.model exists.
+	manifest.model ??= { fallback: [] };
 
 	// Load environment config
 	const envConfig = await loadEnvConfig(agentDir, envFlag);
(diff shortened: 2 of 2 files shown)


shreyas-lyzr (approved):
Clean implementation of zero-config model auto-detection. The fallback chain is well-ordered, the normalization fix for `manifest.model` is correct and needed, and the build + all 27 tests pass.

One thing worth knowing: `getEnvApiKey('github-copilot')` in `pi-ai` resolves against `COPILOT_GITHUB_TOKEN`, `GH_TOKEN`, and `GITHUB_TOKEN`. Since gitagent itself requires a GitHub token, `github-copilot` will match on any machine where no other LLM provider key is set. The PR already places it last in the priority list, so it only fires as a last resort — but it means a user with only `GITHUB_TOKEN` set and no actual Copilot access will see `github-copilot` auto-selected and then get an auth error at call time rather than a clean config error. Worth a note in the README or error message so it's not confusing.

Minor: `claude-sonnet-4-6` (dash) for anthropic/bedrock vs `claude-sonnet-4.6` (dot) for github-copilot — looked correct against the `pi-ai` model registry, so not a bug.

No tests for `autoDetectModel()` itself — the function is simple and pure enough that the risk is low, but a few unit tests covering key detection order and the undefined-when-no-keys path would be a useful addition.

PR #82, opened 2026-07-21
fix(sdk): add Query.continue() to resume after a failed model turn

amish-pratap-singh-lyzr:
## Summary

- `query()` had no way to recover when the model failed to return output. When a turn ended with `stopReason: "error"` (or `"aborted"`), the SDK just emitted a system error message and ended the session — no retry, no continuation.
- `Query.steer()` was also a no-op stub (`steer(_message) {}`), despite the underlying `@mariozechner/pi-agent-core` `Agent` already exposing working `steer()`/`followUp()`/`continue()` primitives that were simply never wired up.

This PR wires both up:

- **`steer(message)`** now actually queues a steering message via `agent.steer(...)` instead of doing nothing. Effective while the agent is still actively running a turn.
- **`continue(message?)`** (new) resumes a session that has gone idle — most importantly after a failed turn:
  - With no argument, it retries the exact turn the model failed to answer: the failed assistant message is dropped so the transcript ends on the user/tool-result message it never responded to, then `agent.continue()` re-sends it (matches `pi-agent-core`'s documented `continue()` contract: "Continue from the current transcript. The last message must be a user or tool-result message.").
  - With a message, it queues a follow-up turn via `agent.followUp(...)` and resumes.
- The internal message channel gained a `reopen()` so a query whose async iterator already reached `agent_end`/`done: true` can resume streaming once `continue()` is called.

## Test plan

- [x] `npm run build` (tsc) — clean, no type errors
- [x] `npm test` — all existing 27 tests pass unchanged; added 4 new tests covering the `continue()`/`steer()` guard paths (method presence, rejecting before the agent has loaded, matching prior `steer()` behavior)
- [x] Manually verified the actual retry mechanics against `pi-agent-core`'s real `Agent` class with a fake `streamFn` that fails once then succeeds — confirmed `continue()` produces a clean transcript (user message + successful assistant reply) rather than leaving the failed message in place

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -1,5 +1,5 @@
 import { Agent } from "@mariozechner/pi-agent-core";
-import type { AgentEvent, AgentTool } from "@mariozechner/pi-agent-core";
+import type { AgentEvent, AgentMessage, AgentTool } from "@mariozechner/pi-agent-core";
 import type { AssistantMessage } from "@mariozechner/pi-ai";
 import { loadAgent } from "./loader.js";
 import type { AgentManifest } from "./loader.js";
@@ -35,6 +35,9 @@ import {
 interface Channel<T> {
 	push(v: T): void;
 	finish(): void;
+	// Undoes finish() so a completed query can resume streaming — used by
+	// Query.continue() to reopen the channel after the run first went idle.
+	reopen(): void;
 	pull(): Promise<IteratorResult<T>>;
 }
 
@@ -59,6 +62,9 @@ function createChannel<T>(): Channel<T> {
 				resolve = null;
 			}
 		},
+		reopen() {
+			done = false;
+		},
 		pull(): Promise<IteratorResult<T>> {
 			if (buffer.length) {
 				return Promise.resolve({ value: buffer.shift()!, done: false });
@@ -72,6 +78,14 @@ function createChannel<T>(): Channel<T> {
 }
 
 // ── Extract text/thinking from AssistantMessage ────────────────────────
+
+function toUserMessage(text: string): AgentMessage {
+	return {
+		role: "user",
+		content: [{ type: "text", text }],
+		timestamp: Date.now(),
+	} as AgentMessage;
+}
 
 function extractContent(msg: AssistantMessage): { text: string; thinking: string } {
 	let text = "";
@@ -111,6 +125,9 @@ export function query(options: QueryOptions): Query {
 	let sandboxCtx: SandboxContext | undefined;
 	// Local session (hoisted for cleanup in catch)
 	let localSession: LocalSession | undefined;
+	// Agent instance (hoisted so steer()/continue() can reach it once the
+	// agent is loaded, including after the initial run has gone idle)
+	let agent: Agent | undefined;
 
 	// OpenTelemetry session span — opened immediately so it covers agent
 	// load + prompt + cleanup. Closed in the IIFE's finally so every exit
@@ -299,7 +316,7 @@ export function query(options: QueryOptions): Query {
 		}
 
 		// 8. Create Agent
-		const agent = new Agent({
+		agent = new Agent({
 			initialState: {
 				systemPrompt,
 				model: loaded.model,
@@ -307,6 +324,9 @@ export function query(options: QueryOptions): Query {
 				...modelOptions,
 			},
 		});
+		// Non-null alias: `agent` is a `let` captured by closures below, so TS
+		// can't narrow it past the assignment above once inside them.
+		const activeAgent = agent;
 
 		// 9. Subscribe to events and map to GCMessage
 		agent.subscribe((event: AgentEvent) => {
@@ -508,7 +528,7 @@ export function query(options: QueryOptions): Query {
 				}
 			}
 			await otelContext.with(_session.ctx, () =>
-				agent.prompt(options.prompt as string),
+				activeAgent.prompt(options.prompt as string),
 			);
 		} else {
 			// Multi-turn: iterate the async iterable
@@ -532,7 +552,7 @@ export function query(options: QueryOptions): Query {
 					}
 				}
 				await otelContext.with(_session.ctx, () =>
-					agent.prompt(userMsg.content),
+					activeAgent.prompt(userMsg.content),
 				);
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/sdk.ts
```diff
@@ -593,7 +613,51 @@ export function query(options: QueryOptions): Query {
...
+			}
+			agent.steer(toUserMessage(message));
+		},
+
+		async continue(message?: string) {
+			if (!agent) {
+				throw new Error("Cannot continue: agent not yet initialized (query hasn't started)");
+			}
+			if (agent.signal) {
+				throw new Error("Cannot continue: agent is still processing a turn");
```
shreyas-lyzr, line 628: Bug 1 (confirmed, fix incomplete): this guard fires a false positive when `continue()` is called from within an `agent_end` event listener — the most natural point to retry after a failure.

The sequence: `runWithLifecycle` emits `agent_end` via `processEvents`, which awaits each subscriber in order. One subscriber calls `channel.finish()`, which synchronously resolves the pending `pull()` promise, scheduling the consumer's for-await continuation as a microtask. That microtask can run before `processEvents` returns, meaning `finishRun()` (which clears `activeRun` and therefore `agent.signal`) hasn't run yet. `agent.signal` is still a live, non-aborted `AbortSignal` — truthy — so this line throws even though the run is effectively done.

Two options:
1. `await agent.waitForIdle()` before the guard — but that changes semantics for a genuinely concurrent call (silent wait instead of immediate throw).
2. Let pi-agent-core's own guard fire (`"Agent is already processing"`), catch it, and re-throw with the friendlier message — skipping `channel.reopen()` on that path.

src/sdk.ts
```diff
@@ -593,7 +613,51 @@ export function query(options: QueryOptions): Query {
...
+
+			// The initial run may have already reached agent_end and finished
+			// the channel — reopen it so further messages can stream out.
+			channel.reopen();
+
+			try {
+				if (message !== undefined) {
+					pushMsg({ type: "user", content: message });
+					agent.followUp(toUserMessage(message));
+				} else {
```
shreyas-lyzr, line 639: Bug 2 (confirmed, triggering condition narrower than stated): `agent.followUp(msg)` queues the message, but pi-agent-core's `agent.continue()` only drains the follow-up queue when the last message role in `state.messages` is `"assistant"`. If the last role is anything else, `continue()` falls through to `runContinuation()`, which ignores the queued follow-up entirely — the new message is silently lost and the LLM re-answers the existing context.

In the current code paths this is handled: `handleRunFailure` unconditionally pushes an error assistant message, so the last role is always `"assistant"` after any failure. The scenario from the PR description — "401 before any assistant message is recorded" — is not reachable via the current `StreamFn` contract (errors surface as a final encoded message, not as a thrown exception). The bug is real but latent; it would resurface if a future pi-agent-core version changes that invariant, or if the agent state is externally modified before `continue()` is called.

src/sdk.ts
```diff
@@ -111,6 +125,9 @@ export function query(options: QueryOptions): Query {
 	let sandboxCtx: SandboxContext | undefined;
 	// Local session (hoisted for cleanup in catch)
 	let localSession: LocalSession | undefined;
+	// Agent instance (hoisted so steer()/continue() can reach it once the
+	// agent is loaded, including after the initial run has gone idle)
+	let agent: Agent | undefined;
 
 	// OpenTelemetry session span — opened immediately so it covers agent
 	// load + prompt + cleanup. Closed in the IIFE's finally so every exit
```
shreyas-lyzr, line 133: Bug 3 (new, moderate): every `continue()` call re-emits a `session_start` system message into the channel. The `agent_start` event fires again at the top of `runWithLifecycle` -> `runAgentLoopContinue` for both paths (with and without a new message), and the subscriber that handles `agent_start` unconditionally emits `{ type: "system", subtype: "session_start" }`. A query that fails and is continued twice will produce three `session_start` and three `session_end` messages within the same `Query` lifetime.

Any consumer that treats `session_start` as a once-per-query lifecycle marker — initialising UI state, starting timers, logging a session — will misfire on every `continue()` call. Consider gating the emit on a flag (e.g. `let sessionStartEmitted = false`) so it only fires once per `Query`.

src/sdk.ts
```diff
@@ -593,7 +613,51 @@ export function query(options: QueryOptions): Query {
...
+			}
+
+			// The initial run may have already reached agent_end and finished
+			// the channel — reopen it so further messages can stream out.
+			channel.reopen();
+
+			try {
+				if (message !== undefined) {
+					pushMsg({ type: "user", content: message });
+					agent.followUp(toUserMessage(message));
```
shreyas-lyzr, line 638: Bug 4 (new, minor but observable): `pushMsg({ type: "user", content: message })` is called before `agent.followUp()` and before `await agent.continue()`. If the subsequent `agent.continue()` call rejects (e.g. pi-agent-core throws `"Cannot continue from message role: assistant"` due to unexpected state), the catch block pushes an error system message and calls `channel.finish()`. The consumer's `messages()` now contains the user message as if it was processed by the LLM, when it never was — a ghost entry that persists in `collectedMessages` for the lifetime of the query.

Consider pushing the user message only after `agent.followUp()` and `agent.continue()` resolve successfully, so the consumer-visible transcript stays consistent with what was actually sent.

src/sdk.ts
```diff
@@ -593,7 +613,51 @@ export function query(options: QueryOptions): Query {
...
+					// pi-agent-core's continue() requires that to retry in place.
+					const messages = agent.state.messages as any[];
+					const last = messages[messages.length - 1];
+					if (last?.role === "assistant" && (last.stopReason === "error" || last.stopReason === "aborted")) {
+						agent.state.messages = messages.slice(0, -1);
+					}
+				}
+
+				await otelContext.with(_session.ctx, () => (agent as Agent).continue());
+			} catch (err: any) {
```
shreyas-lyzr, line 652: Bug 5 (new, minor): `_session.end()` is called in the IIFE's `finally` block, which runs when the first run ends — before any `continue()` call. Child spans created during a continued run (gen_ai.chat spans, tool spans) are created under `_session.ctx` via `otelContext.with(_session.ctx, ...)` here, but by this point the parent session span is already ended. This produces broken telemetry traces with out-of-order or orphaned child spans.

The session span should remain open until the `Query` itself is done. One approach: move `_session.end()` into a `finalise()` helper called from both the normal `channel.finish()` path and the `return()`/`throw()` methods of the generator, guarded by a flag so it fires exactly once.

shreyas-lyzr (changes requested):
Five issues found — two confirming the bugs noted in the PR description (one with a more precise failure path), three new ones. Inline comments below.

PR #86, opened 2026-07-23
feat: add Plivo SMS and voice skills

### LINKED ISSUE #83: feat: Add Plivo SMS and voice skills
## Summary

Add two skills, `plivo-sms` and `plivo-voice`, that let the agent send SMS and place outbound voice calls via Plivo, following the existing `gmail-email` skill pattern (a skill dir with `SKILL.md` + `scripts/`).

## Motivation

The agent can reach people over email (`gmail-email`) but can't text or call. SMS and voice let it notify the user when it finishes a task, hits a blocker, or a hook fires — and reach the user when they are away from the terminal, where a text or a call lands better than email.

## Proposed Solution

Two new skills under `skills/`, mirroring `gmail-email`:
- `skills/plivo-sms/` — `SKILL.md` + `scripts/send_sms.py` (POST to the Plivo Messages API).
- `skills/plivo-voice/` — `SKILL.md` + `scripts/make_call.py` (POST to the Plivo Call API — the answer XML speaks the message).

Auth is HTTP Basic (Auth ID + Auth Token) via env vars / `.env`. Scripts are stdlib-only (no new dependencies). Both are auto-discovered by `discoverSkills()` in `src/skills.ts` (it reads `skills/`), so no registration is needed.

## Scope

Outbound only (send SMS, place call). Inbound isn't included as the core has no webhook server, so receiving replies is out of scope here.

## Alternatives

Hand-rolling the Plivo REST calls in a one-off script works, but isn't a discoverable, reusable skill and doesn't fit the `skills/` pattern.

## Additional Context

MIT-compatible, no new dependencies. Implementation is ready and can open a PR. Happy to adjust to your conventions.

sarveshpatil-plivo:
Closes #83

## What

Adds two skills, `plivo-sms` and `plivo-voice`, so an agent can text or call someone through Plivo. They sit next to the existing `gmail-email` skill and follow the same layout, a `SKILL.md` with YAML frontmatter and a script under `scripts/`.

## How it works

Both scripts call the Plivo REST API with HTTP Basic auth, reading `PLIVO_AUTH_ID` and `PLIVO_AUTH_TOKEN` from the environment or a `.env` file in the skill directory. They use only the Python standard library, so there is nothing extra to install, same as gmail-email. Both are picked up automatically by `discoverSkills()` in `src/skills.ts`, so no registration is needed.

- `plivo-sms` sends an SMS through the Messages API. Arguments are `--from`, `--to`, `--text`.
- `plivo-voice` places an outbound call through the Call API. Arguments are `--from`, `--to`, `--answer-url`. When the call connects Plivo runs the XML at the answer URL, which can speak a message with `<Speak>`.

Files added

- `skills/plivo-sms/SKILL.md`
- `skills/plivo-sms/scripts/send_sms.py`
- `skills/plivo-voice/SKILL.md`
- `skills/plivo-voice/scripts/make_call.py`

## Scope

Outbound only, matching the issue. Inbound is left out since the core has no webhook server.

## Testing

Ran the SMS skill against a live Plivo account and the message was delivered. The voice skill uses the same auth and request pattern against the Call API. Credentials are read from the environment, so nothing sensitive is in the code.

### DIFF
--- a/skills/plivo-sms/SKILL.md
+++ b/skills/plivo-sms/SKILL.md
@@ -1,0 +1,50 @@
+---
+name: plivo-sms
+description: Send SMS messages via the Plivo Messages API using Auth ID/Auth Token authentication.
+---
+
+# Plivo SMS Skill
+
+Send SMS messages via Plivo.
+
+## Setup
+
+1. **Create a Plivo account** and open the console at https://cx.plivo.com
+2. **Get your credentials**:
+   - Copy your **Auth ID** and **Auth Token** from the dashboard
+   - Rent a Plivo phone number (or use an approved sender ID) to send from
+
+3. **Configure credentials**:
+   ```bash
+   export PLIVO_AUTH_ID="your-auth-id"
+   export PLIVO_AUTH_TOKEN="your-auth-token"
+   ```
+
+   Or create a `.env` file in the skill directory:
+   ```
+   PLIVO_AUTH_ID=your-auth-id
+   PLIVO_AUTH_TOKEN=your-auth-token
+   ```
+
+## Usage
+
+```bash
+python3 scripts/send_sms.py \
+  --from "+14150000002" \
+  --to "+14150000001" \
+  --text "Message body text"
+```
+
+Send to multiple recipients by joining numbers with `<`:
+
+```bash
+python3 scripts/send_sms.py \
+  --from "+14150000002" \
+  --to "+14150000001<+14160000003" \
+  --text "Message body text"
+```
+
+## Requirements
+
+- Python 3.6+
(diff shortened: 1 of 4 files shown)


Nivesh353:
@sarveshpatil-plivo  Thanks for the contribution! Confirmed these follow the same pattern as the existing Gmail email skill: just a `SKILL.md` and script, with no tool registration or automatic wiring into the codebase. They only become available when copied into a specific agent's `skills/` directory, and even then, the model must explicitly choose to invoke the script based on the skill instructions.

For context, we already have Twilio integrated for calls and SMS in `--voice` mode.

Closing this PR since it doesn't add functionality that's integrated into GitAgent itself.

PR #88, opened 2026-07-26
fix(repo-mode): prevent unsafe dir defaults and forced pushes (#85)

### LINKED ISSUE #85: repo: mode: unsafe dir default can reset an unrelated parent git repo; unconditional push on every call with no read-only mode
query()'s repo: option has two related problems that make it unsafe to use from anything other than an interactive one-shot CLI session: (1) dir is typed as required but the runtime silently defaults to process.cwd(), which can resolve to an unrelated repo's working tree and get git reset --hard; (2) finalize() (commit + push) runs unconditionally at the end of every call, success or failure, with no way to opt out.

Bug 1 — LocalRepoOptions.dir type says required, runtime treats it as optional and defaults to process.cwd()
session.d.ts:

export interface LocalRepoOptions {
  url: string;
  token: string;
  dir: string;      // <- no `?`, looks mandatory
  session?: string;
}
sdk.js:

let dir = options.dir ?? process.cwd();
if (options.repo) {
  ...
  localSession = initLocalSession({
    url: options.repo.url,
    token,
    dir: options.repo.dir || dir,   // <- silently falls back to cwd if unset
    session: options.repo.session,
  });
If the calling process's cwd has no .git of its own (e.g. it's a subdirectory like frontend/ inside a monorepo), git's normal directory-walk-up behavior means every command initLocalSession runs (checkout, reset --hard) executes against whatever parent repo git finds — which may have nothing to do with the repo.url being cloned.

Reproduction:

From a directory with no .git of its own, but nested inside a parent repo (e.g. some-monorepo/apps/web/), call:

query({ prompt: "hi", repo: { url: "https://github.com/org/some-other-repo", token } })
(no dir passed)
initLocalSession finds some-monorepo/apps/web already "exists" (it's a real directory), so it takes the update branch: git remote set-url origin <authed-url>, git fetch, git checkout <default>, git reset --hard origin/<default> — all run with cwd: some-monorepo/apps/web.
Because that directory has no .git, these commands resolve to some-monorepo's actual .git, rewriting some-monorepo's origin remote (with the token embedded in the URL) and attempting to hard-reset the entire monorepo onto some-other-repo's content.
In our case this happened three times in one session; it was only prevented from actually destroying data because git checkout refused to proceed due to uncommitted local changes — that's git's own safety net, not gitagent's. A directory with committed state, or a clean tree, would not have been protected.

Suggested fix: Make dir actually required at the type level and at runtime (throw if options.repo.dir is unset — don't fall back to process.cwd()). At minimum, before running any destructive git command, verify the resolved dir either doesn't exist yet (safe to clone into) or its own git rev-parse --show-toplevel equals dir itself (i.e., it's not inside a different repo's working tree).

Bug 2 — finalize() (commit + push) is unconditional, no read-only mode
In sdk.js, localSession.finalize() is called at the end of every successful run and in the top-level .catch() handler on error — there's no option to skip the commit/push step:

// success path
if (localSession) {
  try { localSession.finalize(); } catch { /* best-effort */ }
}
...
// error path
.catch(async (err) => {
  if (localSession) {
    try { localSession.finalize(); } catch { /* best-effort */ }
  }
  ...
finalize() → commitChanges() (no-op if nothing staged) then unconditionally push(). When session isn't passed, initLocalSession mints a fresh random branch (gitagent/session-<8-hex>) every call — meaning every single query() invocation with repo: set pushes a new branch to the real remote, even if the model made zero changes (tools disabled, pure Q&A). There's no way to run repo: mode read-only.

Suggested fix: Add an option like repo.readOnly or repo.autoPush: false that skips finalize()/push() entirely, for callers that only want gitagent to read an agent definition from a repo without mutating it.

Why this matters
The combination of these two makes repo: unsafe to wire into any server/request path (as opposed to a one-shot interactive CLI session) — which is a reasonable thing to want, since the SDK is marketed as embeddable ("mirrors the Claude Agent SDK pattern but runs in-process"). Right now the only safe pattern is: never omit dir, always pass a pinned session, and accept that every call still pushes to the remote regardless of intent. PS: Its wiped out my frontend..

Nivesh353:
## Summary
- Repo mode (`repo:` in the SDK, `--repo` in the CLI) silently defaulted the
  working directory to `process.cwd()` when `dir` was omitted. If that
  directory had no `.git` of its own, git commands would escape upward and
  run destructive operations (`reset --hard`, `remote set-url`) against
  whatever ancestor repo was found instead — corrupting an unrelated project.
- `finalize()` always committed and pushed on every call, success or failure,
  with no way to opt out — even for pure read-only use.
- Fixes both, closes #85:
  - `dir` is now required in repo mode; omitting it throws instead of
    silently falling back to cwd.
  - Before touching an existing directory, `initLocalSession` now verifies
    it's a git repo root in its own right (`git rev-parse --show-toplevel`,
    compared via `realpathSync` to handle symlinked paths like macOS's
    `/tmp` → `/private/tmp`) and that it already tracks the expected
    remote — refusing to proceed otherwise.
  - Added `repo.readOnly` (SDK) and `--read-only` (CLI), which skip
    `commitChanges()`/`push()` in `finalize()` while still stripping the
    embedded token from `origin`.
- Also fixed two stale `test/mcp.test.ts` assertions that were checking
  `execute()` against an old plain-string-return contract; the rest of the
  codebase (e.g. `sdk.test.ts`) already expects the current structured
  `{ content, details }` shape from `buildTool()`. Unrelated to #85, found
  while confirming this branch's `npm test` was clean.

## Test plan
- [x] `npm run build` — no type errors
- [x] `npm test` — 39/39 passing (was 37/39 on unmodified `main`; the 2
      failures were the stale `mcp.test.ts` assertions above, now fixed)
- [x] Manually reproduced the original vulnerability (git-less subfolder
      inside a real repo, pointed at a different real repo) — confirmed it
      threw and left the victim repo untouched, instead of resetting it
- [x] Ran repo mode 3x with `readOnly: true` (SDK) and `--read-only` (CLI)
      against a real public repo — 0 pushes landed remotely, confirmed via
      `git ls-remote`
- [x] Ran repo mode without `readOnly` — confirmed normal commit/push
      behavior is unchanged (no regression)

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,5 +1,5 @@
 import { execSync } from "child_process";
-import { existsSync, mkdirSync, writeFileSync } from "fs";
+import { existsSync, mkdirSync, writeFileSync, realpathSync } from "fs";
 import { resolve } from "path";
 import { randomBytes } from "crypto";
 
@@ -10,6 +10,7 @@ export interface LocalRepoOptions {
 	token: string;
 	dir: string;
 	session?: string;
+	readOnly?: boolean;
 }
 
 export interface LocalSession {
@@ -56,6 +57,9 @@ function getDefaultBranch(cwd: string): string {
 
 export function initLocalSession(opts: LocalRepoOptions): LocalSession {
 	const { url, token, session } = opts;
+	if (!opts.dir) {
+		throw new Error("repo.dir is required — refusing to guess a working directory for repo mode");
+	}
 	const dir = resolve(opts.dir);
 	const aUrl = authedUrl(url, token);
 
@@ -63,6 +67,35 @@ export function initLocalSession(opts: LocalRepoOptions): LocalSession {
 	if (!existsSync(dir)) {
 		execSync(`git clone --depth 1 --no-single-branch ${aUrl} ${dir}`, { stdio: "pipe" });
 	} else {
+		// Refuse to operate unless `dir` is a git repo root in its own right.
+		// If `dir` has no local .git, git commands here would silently resolve
+		// to whatever ancestor repo does have one — e.g. a monorepo root.
+		let topLevel: string;
+		try {
+			topLevel = execSync("git rev-parse --show-toplevel", { cwd: dir, stdio: "pipe", encoding: "utf-8" }).trim();
+		} catch {
+			throw new Error(`${dir} exists but is not a git repository — refusing to operate on it`);
+		}
+		if (realpathSync(topLevel) !== realpathSync(dir)) {
+			throw new Error(
+				`${dir} has no .git of its own — git commands here would escape to the ancestor repo at ${topLevel}. Refusing to run destructive commands.`,
+			);
+		}
+
+		// Refuse to reset a repo that's already tracking a different remote —
+		// it isn't the clone this call expects, even if the folder exists.
+		let existingOrigin = "";
+		try {
+			existingOrigin = execSync("git remote get-url origin", { cwd: dir, stdio: "pipe", encoding: "utf-8" }).trim();
+		} catch {
+			// no origin configured yet — fine, we're about to set one
+		}
+		if (existingOrigin && cleanUrl(existingOrigin) !== cleanUrl(url)) {
+			throw new Error(
+				`${dir} already tracks a different remote (${cleanUrl(existingOrigin)}), not ${url}. Refusing to overwrite it.`,
+			);
+		}
+
 		git(`remote set-url origin ${aUrl}`, dir);
 		git("fetch origin", dir);
 
@@ -144,9 +177,11 @@ export function initLocalSession(opts: LocalRepoOptions): LocalSession {
 		},
 
 		finalize() {
-			localSession.commitChanges();
-			localSession.push();
-			// Strip PAT from remote URL
+			if (!opts.readOnly) {
+				localSession.commitChanges();
+				localSession.push();
+			}
+			// Strip PAT from remote URL regardless of readOnly
 			git(`remote set-url origin ${cleanUrl(url)}`, dir);
 		},
 	};
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,4 +1,4 @@
-import { execSync } from "child_process";
+import { execSync, execFileSync } from "child_process";
 import { existsSync, mkdirSync, writeFileSync, realpathSync } from "fs";
 import { resolve } from "path";
 import { randomBytes } from "crypto";
@@ -30,7 +30,7 @@ function authedUrl(url: string, token: string): string {
 }
 
 function cleanUrl(url: string): string {
-	return url.replace(/^https:\/\/[^@]+@/, "https://");
+	return url.replace(/^https:\/\/[^@]+@/, "https://").replace(/\.git$/, "");
 }
 
 function git(args: string, cwd: string): string {
@@ -96,7 +96,10 @@ export function initLocalSession(opts: LocalRepoOptions): LocalSession {
 			);
 		}
 
-		git(`remote set-url origin ${aUrl}`, dir);
+		// `set-url` only updates an existing remote — it errors if `origin`
+		// isn't configured yet, which happens if `dir` is a repo the caller
+		// created themselves rather than one gitagent cloned previously.
+		git(`remote ${existingOrigin ? "set-url" : "add"} origin ${aUrl}`, dir);
 		git("fetch origin", dir);
 
 		// Reset local default branch to latest remote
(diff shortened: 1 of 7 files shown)


src/session.ts
```diff
@@ -56,13 +57,45 @@ function getDefaultBranch(cwd: string): string {
...
+		// Refuse to reset a repo that's already tracking a different remote —
+		// it isn't the clone this call expects, even if the folder exists.
+		let existingOrigin = "";
+		try {
+			existingOrigin = execSync("git remote get-url origin", { cwd: dir, stdio: "pipe", encoding: "utf-8" }).trim();
+		} catch {
+			// no origin configured yet — fine, we're about to set one
+		}
+		if (existingOrigin && cleanUrl(existingOrigin) !== cleanUrl(url)) {
+			throw new Error(
```
shreyas-lyzr, line 94: This comparison can produce a false positive. `cleanUrl` strips the embedded token but does not normalize the `.git` suffix. A repo cloned via `https://github.com/org/repo.git` stores that URL as its origin. If the caller passes `https://github.com/org/repo`, the two cleaned URLs differ and this throws — even though both refer to the same repo. Add `.replace(/\.git$/, '')` to `cleanUrl` to handle this case.

shreyas-lyzr (changes requested):
Three issues worth addressing before merge — one is a URL normalization bug causing false-positive rejections, one is a latent shell injection in commitChanges, and one is a minor CLI/SDK behavioral inconsistency. The readOnly feature and the git-root validation are solid.

**1. `cleanUrl` does not normalize `.git` suffixes — remote-mismatch check produces false positives**

A repo cloned via `https://github.com/org/repo.git` stores that URL as its origin. If the caller then passes `https://github.com/org/repo` (without `.git`), `cleanUrl(existingOrigin) !== cleanUrl(url)` evaluates to true and the session is refused with a misleading error, even though both point to the same repo. This is a regression vs. the old code path which would overwrite and proceed.

Fix: normalize `.git` in `cleanUrl`.

**2. Shell injection via commit message in `git commit -m "${commitMsg}"`**

`commitChanges` interpolates `commitMsg` directly into a shell command string passed to `execSync`. The default value is safe, but the function accepts arbitrary caller-supplied messages — a message containing a double-quote or `$()` breaks out of the shell argument. Use `execFileSync` with an argument array for this call, or shell-escape the message before interpolation.

**3. CLI `--dir` guard is fragile compared to SDK enforcement**

The SDK throws when `dir` is omitted in repo mode. The CLI silently derives `/tmp/gitagent/<repo-name>` when `--dir` is not passed. The guard used is `dir === process.cwd()` — if the user explicitly passes `--dir $(pwd)`, the guard is bypassed and their actual cwd is used as the repo dir. Compare `rawDir === undefined` instead.

Security pass: no new dependencies in this PR, no hardcoded secrets, no obvious injection/authz gaps beyond finding #2 above.

src/session.ts
Nivesh353: Fixed — cleanUrl now strips the trailing .git suffix before comparing, so url vs url.git no longer false-positives.

shreyas-lyzr (approved):
All three findings from the previous review are addressed. The new 'no origin' fix also looks correct. Here's what I verified:

**Finding 1 — cleanUrl missing .git strip:** Fixed. cleanUrl now strips both the auth prefix and .git suffix (session.ts line 33). The remote-URL comparison at line 93 will correctly match whether or not the caller supplied a .git suffix.

**Finding 2 — shell injection in commitChanges:** Fixed. execFileSync with an argv array is used at line 177. The commit message is passed as a discrete argument rather than shell-interpolated, so arbitrary content is safe.

**Finding 3 — fragile rawDir guard:** Fixed. index.ts now tracks rawDir as undefined when --dir is not passed (parseArgs leaves it unset), and the default-dir branch checks rawDir === undefined (line 354) rather than comparing the string value. Explicit --dir matching cwd is now correctly honoured.

**New fix — no-origin crash:** Correct. When git remote get-url origin fails (empty repo pointed at this URL for the first time), existingOrigin is set to '' and the ternary at line 102 calls git remote add origin ... instead of git remote set-url origin ..., which would have thrown. The subsequent fetch and reset proceed normally. The finalize path (line 191) always calls set-url, which is safe because by that point origin is guaranteed to exist (either from clone, from add, or from the previous set-url).

One pre-existing note (not introduced by this PR, not blocking): the git() helper interpolates aUrl and cleanUrl(url) into a shell string, so a maliciously crafted repo URL with shell metacharacters could inject commands. That pattern exists in main already; it is out of scope here but worth a follow-up issue.

Security pass: no new dependencies, no hardcoded secrets, no new injection surface, no auth/authz gaps. The aUrl interpolation concern pre-dates this branch.

Tests: 39/39 per the PR description; the two previously failing mcp.test.ts assertions are fixed to match the current structured return shape.

PR #90, opened 2026-07-28
feat: run SkillFlows from the CLI, SDK and scheduler

### LINKED ISSUE #89: SkillFlow execution lives only in voice — CLI, SDK, and scheduler can't run flows
Core owns the SkillFlow format — discoverWorkflows, loadFlowDefinition, saveFlowDefinition, deleteFlowDefinition, plus the system-prompt block — but the executor lives only in @open-gitagent/voice (server.ts, executeFlow). Nothing in this repo can run a flow.

Bug this causes: [workflows.ts:156](https://github.com/open-gitagent/gitagent/blob/main/src/workflows.ts#L156) appends "SkillFlows can be triggered with @flow_name in chat" to every agent's system prompt ([loader.ts:313](https://github.com/open-gitagent/gitagent/blob/main/src/loader.ts#L313) → [index.ts:596](https://github.com/open-gitagent/gitagent/blob/main/src/index.ts#L596)). The CLI has no @flow handler — its commands are /memory, /skills, /tasks, /learned, /plugins, /skill:name, /quit, /exit. The model is told a trigger exists that doesn't, and typing it falls through as ordinary chat.

Repro: run the CLI in any agent dir with a workflows/*.yaml containing steps: — the sentence is in the system prompt.

Also blocked:

Scheduler holds a single prompt string, so a flow named daily-report cannot run daily.
SDK users can't build on flows at all — query() has no flow concept.
Proposal: move the step loop into core with its side effects injected:

executeFlow(flow, input, { runStep, onProgress, requestApproval })
voice passes its WebSocket broadcaster and Telegram/WhatsApp approver, CLI passes console.log and a terminal y/n, scheduler passes a logger and an explicit approval policy. The visual builder stays in voice — only the loop moves.

Nivesh353:
Closes #89
## Problem
Core owns the SkillFlow format — discoverWorkflows, loadFlowDefinition, saveFlowDefinition, deleteFlowDefinition, and the workflows block injected into every system prompt. But the executor lived only in @open-gitagent/voice. So the CLI, SDK and scheduler could create, list and delete flows without ever being able to run one — while [workflows.ts:156](https://github.com/open-gitagent/gitagent/blob/main/src/workflows.ts#L156) told every agent, including the CLI's, that @flow_name worked.

## Approach
Move the step loop into core as src/flow-runner.ts, with its side effects injected:

executeFlow(flow, input, { runStep, onProgress, requestApproval })
The loop owns ordering, prompt construction, {input} substitution, context threading and gate semantics. Each caller supplies only what's genuinely specific to it — voice passes its WebSocket broadcaster and Telegram/WhatsApp approver, the CLI passes console.log and a readline prompt, the scheduler passes a logger. The visual builder stays in voice; only the loop moved.

## What's included
src/flow-runner.ts — the loop, plus a typed event stream (flow_start, step_start, step_done, approval_requested, approval_resolved, flow_done, flow_aborted) that front-ends render however they like.

src/run-flow.ts — runFlow(), a wrapper supplying the obvious default (each step is an isolated query() against the agent dir) so callers don't hand-write a runStep, plus per-step token usage:

const r = await runFlow({ agentDir, flow: "daily-report", input: "..." });
CLI — @flow-name [input] trigger and a /flows command. The trigger is anchored to line start, so an address like bob@daily-report.com can't fire a flow. Gates prompt on the existing readline instance. Both re-discover workflows rather than using the startup snapshot, matching how /skills calls refreshSkills.

Schedules — new optional flow field, so a schedule can run a SkillFlow instead of a single prompt (prompt becomes the flow's input). A flow schedule with no executor fails loudly rather than silently doing nothing.

## Behaviour change worth reviewing
Approval gates now deny when no handler is supplied. A gate guards a risky step; with nobody to ask, continuing silently removes the protection it exists to provide. Unattended callers opt in explicitly with approve: "auto".

This changes voice's current behaviour, which auto-approves when no channel is connected — addressed in the companion voice PR.

## Testing
21 new tests, none requiring an API key — gate-only flows resolve before any model call. Also verified end to end against a real agent:

CLI: /flows, @flow, gate approved and denied, unknown flow, and the anchored trigger correctly ignoring bob@daily-report.com
SDK: executeFlow with a custom runStep; runFlow across deny / "auto" / a human approver
A scheduled flow firing on cron and stopping at its gate
Per-step token accounting reconciled against billing to the last digit

### DIFF
--- a/src/flow-runner.ts
+++ b/src/flow-runner.ts
@@ -1,0 +1,132 @@
+// Executes a SkillFlow: runs each step in order, threading every step's output
+// into the next one's prompt. All side effects are injected (RunFlowDeps) so the
+// same loop drives the CLI, the scheduler, the SDK, and the voice web UI —
+// the caller supplies how to run a step, how to report progress, and how to ask
+// a human for approval.
+
+import type { SkillFlowDefinition, SkillFlowStep } from "./workflows.js";
+
+/** Reserved pseudo-skill: pauses the flow until a human approves. */
+export const APPROVAL_GATE = "__approval_gate__";
+
+export type FlowEvent =
+	| { type: "flow_start"; flow: string; totalSteps: number }
+	| { type: "step_start"; index: number; total: number; skill: string }
+	| { type: "step_done"; index: number; total: number; skill: string; output: string }
+	| { type: "approval_requested"; index: number; total: number; channel?: string }
+	| { type: "approval_resolved"; index: number; approved: boolean }
+	| { type: "flow_done"; flow: string; steps: number }
+	| { type: "flow_aborted"; flow: string; index: number; reason: string };
+
+export interface FlowRunDeps {
+	/**
+	 * Run one step and return the assistant's text. Receives the fully-built
+	 * prompt plus the raw step and its index — so a caller can inspect the step
+	 * for per-step settings (e.g. a future `model` field) without the loop
+	 * needing to know about them.
+	 */
+	runStep: (prompt: string, step: SkillFlowStep, index: number) => Promise<string>;
+	/** Progress reporting — terminal output, WebSocket frames, log lines. */
+	onProgress?: (event: FlowEvent) => void;
+	/**
+	 * Ask a human to approve continuing. Omit it and approval gates DENY:
+	 * a gate exists to guard a risky step, so with nobody to ask, the safe
+	 * answer is to stop. Callers wanting unattended runs must opt in explicitly
+	 * by passing `async () => true`.
+	 */
+	requestApproval?: (message: string, channel?: string) => Promise<boolean>;
+}
+
+export interface FlowStepResult {
+	index: number;
+	skill: string;
+	output: string;
+}
+
+export interface FlowResult {
+	flow: string;
+	/** False when an approval gate denied, timed out, or had no handler. */
+	completed: boolean;
+	steps: FlowStepResult[];
+	/** The accumulated context after the last step that ran. */
+	context: string;
+	/** 0-based index of the step that stopped the flow, when not completed. */
+	abortedAt?: number;
+	abortReason?: string;
+}
+
+function buildStepPrompt(step: SkillFlowStep, userContext: string, runningContext: string): string {
+	return `Use the skill "${step.skill}" (load it with /skill:${step.skill}).
+${step.prompt.replace(/\{input\}/g, userContext)}
+
+Context from previous steps:
+${runningContext}`;
+}
+
+function buildApprovalMessage(
+	flow: SkillFlowDefinition,
+	step: SkillFlowStep,
+	index: number,
+	runningContext: string,
+): string {
+	if (step.prompt) {
+		return `⏸ Approval Required: ${step.prompt}\n\nReply YES to continue or NO to cancel.`;
+	}
+	return `⏸ Flow "${flow.name}" paused at step ${index + 1}/${flow.steps.length}.\n\nCompleted so far:\n${runningContext.slice(0, 500)}\n\nReply YES to continue or NO to cancel.`;
+}
+
+/**
+ * Run a SkillFlow to completion, or stop at the first denied approval gate.
+ * Never throws on a denied gate — that's a normal outcome, reported via
+ * `completed: false`. Errors from `runStep` propagate to the caller.
+ */
+export async function executeFlow(
+	flow: SkillFlowDefinition,
+	userContext: string,
+	deps: FlowRunDeps,
+): Promise<FlowResult> {
+	const total = flow.steps.length;
+	const results: FlowStepResult[] = [];
+	let runningContext = userContext;
+
+	deps.onProgress?.({ type: "flow_start", flow: flow.name, totalSteps: total });
+
+	for (let i = 0; i < total; i++) {
+		const step = flow.steps[i];
+
+		if (step.skill === APPROVAL_GATE) {
+			const channel = step.channel;
+			deps.onProgress?.({ type: "approval_requested", index: i, total, channel });
+
+			// No handler → deny. See RunFlowDeps.requestApproval.
+			const approved = deps.requestApproval
+				? await deps.requestApproval(buildApprovalMessage(flow, step, i, runningContext), channel)
+				: false;
+
+			deps.onProgress?.({ type: "approval_resolved", index: i, approved });
+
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -788,6 +861,19 @@ async function main(): Promise<void> {
 				return;
 			}
 
+			// SkillFlow trigger: @flow-name [input]. Anchored to the start of the
+			// line so an @mention or email address mid-sentence can't fire a flow.
+			const flowMatch = trimmed.match(/^@([a-z0-9]+(?:-[a-z0-9]+)*)\b\s*([\s\S]*)$/);
```
shreyas-lyzr: The `\b` word-boundary anchor doesn't fully prevent unintended matches. `@daily-report.com` matches with `flowName=daily-report` and `input='.com'`, because `\b` sits between the word char `t` and the non-word char `.`.

The PR description claims this is guarded, but it isn't — only `bob@daily-report.com` (non-start) is blocked.

Fix: require the flow name to be followed by whitespace or end-of-string rather than any word boundary:

```suggestion
			const flowMatch = trimmed.match(/^@([a-z0-9]+(?:-[a-z0-9]+)*)(?=\s|$)\s*([\s\S]*)$/);
```

This rejects `@daily-report.com` entirely rather than firing flow `daily-report` with input `.com`.

src/run-flow.ts
```diff
@@ -0,0 +1,157 @@
...
+	// "deny" and undefined both mean: no handler, so executeFlow denies and
+	// says why. Returning undefined keeps that message accurate.
+	return undefined;
+}
+
+async function resolveFlow(agentDir: string, flow: string | SkillFlowDefinition): Promise<SkillFlowDefinition> {
+	if (typeof flow !== "string") return flow;
+
+	const workflows = await discoverWorkflows(agentDir);
+	const meta = workflows.find((w) => w.name === flow);
```
shreyas-lyzr: `resolveFlow` finds the first workflow matching by name regardless of type, then errors if it turns out to be a reference. If `daily-report.md` and `daily-report.yaml` both exist in the `workflows/` directory, `readdir` order (filesystem-dependent) determines which one `find` returns — and if the markdown file comes first, the caller gets a confusing "reference workflow, not runnable" error even though the flow exists.

The CLI path in `index.ts` already handles this correctly with `w.name === flowName && w.type === "flow"`. Apply the same filter here:

```suggestion
	const meta = workflows.find((w) => w.name === flow && w.type === "flow");
```

Then simplify the post-find type check — it becomes unreachable and can be removed.

src/flow-runner.ts
```diff
@@ -0,0 +1,132 @@
...
+	abortedAt?: number;
+	abortReason?: string;
+}
+
+function buildStepPrompt(step: SkillFlowStep, userContext: string, runningContext: string): string {
+	return `Use the skill "${step.skill}" (load it with /skill:${step.skill}).
+${step.prompt.replace(/\{input\}/g, userContext)}
+
+Context from previous steps:
+${runningContext}`;
```
shreyas-lyzr, line 63: For the first step (index 0), `runningContext` is initialised to `userContext`, so the step's prompt includes the user's input twice: once substituted into `{input}` inside the step prompt, and once verbatim in the "Context from previous steps" block.

This is wasteful for large inputs and adds noise the model has to filter out. For step 0, the context block should reflect that there are no previous steps yet, e.g. passing an empty string or a neutral label:

```suggestion
	let runningContext = "";
```

`{input}` substitution in the prompt already covers the user's text for step 0; the context block only needs previous step outputs.

shreyas-lyzr (changes requested):
Three issues to address before merge — one is a correctness bug that affects real-world inputs, one is a correctness inconsistency between the CLI and SDK paths, and one is a cosmetic/waste issue worth cleaning up. The overall structure is solid: the loop, injection model, and approval gate semantics are well-designed, and the test coverage is thorough.

Security pass: no new dependencies added, no secrets in the diff, no obvious injection or authz gaps. The approval-gate-denies-by-default change is the right call.

src/index.ts
Nivesh353: \b regex. Confirmed: @daily-report.com did fire the flow with input .com. Applied your suggestion verbatim. Verified @daily-report.com → no match, @daily-report hello → fires with input hello, @daily-report → fires with empty input. The same line existed in the companion voice PR; fixed there too.

src/run-flow.ts
Nivesh353: resolveFlow type filter. Real bug, thanks. I've kept the post-find type check rather than removing it — filtering by type === "flow" alone makes meta undefined when only a .md exists, so you'd get "not found" instead of the more useful "that's a reference workflow", and it fails the existing markdown test. Used preferred-then-fallback so both cases work:
const meta = workflows.find((w) => w.name === flow && w.type === "flow")
  ?? workflows.find((w) => w.name === flow);
Added a test that creates both gate-only.md and gate-only.yaml and asserts the YAML resolves.

src/flow-runner.ts
Nivesh353: I'd rather not change this. runningContext = "" silently drops the user's input for any flow whose first step doesn't template {input} — the context block is the only place it appears. That's 4 of the 5 flows in my test agent (frontend-review, risky-flow, safe-flow, trusted-flow); @frontend-review build me a landing page would reach the agent with no mention of a landing page. It's also verbatim voice's existing behaviour, and this PR is otherwise a faithful port.

You're right that something's off, but I think it's the label — "Context from previous steps" is describing user input on step 0, when there are no previous steps. Happy to relabel it here or in a follow-up; emptying it trades a duplicated string for lost input.

Tests now 22/22, both repos typecheck clean.

shreyas-lyzr (approved):
The follow-up commit addresses all three issues cleanly.

Regex anchor: `(?=\s|$)` replaces the word boundary — `@daily-report.com` no longer matches. The fix is also applied in the voice server.

resolveFlow: prefers a `type === "flow"` match before falling back to any match. The index.ts REPL path already filtered by type and is consistent.

Context doubling: the author made a deliberate call not to change the initial `runningContext = userContext` — correctly noted that emptying it would drop the input for flows whose first step doesn't template `{input}`. That's a reasonable trade-off and within scope for the author to own.

Security pass: no new dependencies, no secrets, no injection or authz gaps.

Looks good to merge.

PR #94, opened 2026-07-28
feat(workflows): expose transport-agnostic flow executor

tomatotomata:
## Summary

- expose a reusable `executeFlow` primitive from core for CLI, SDK, scheduler, and voice adapters
- thread each step's output into the next step
- support `__approval_gate__` through an injected callback with explicit rejected status
- emit transport-neutral progress events and add focused coverage

This addresses the missing core execution loop described in #89 without coupling the workflow module to a specific transport or agent runtime.

## Verification

- `npm run build`
- `npm test` (68 tests)
- `node --test test/workflows.test.ts --experimental-strip-types` (3 tests)
- `git diff --check`

### DIFF
--- a/src/exports.ts
+++ b/src/exports.ts
@@ -86,9 +86,18 @@ export { getVoiceContext, getAgentContext } from "./context.js";
 export { discoverSkills } from "./skills.js";
 export {
 	discoverWorkflows,
+	executeFlow,
 	loadFlowDefinition,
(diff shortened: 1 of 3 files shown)


tomatotomata:
Closing as superseded by the existing upstream PR #90, which covers the same SkillFlow execution surface. The implementation is retained locally for reference.

PR #102, opened 2026-08-07
feat(skills): outcome-driven skill repair, with a human in the loop

Nivesh353:
## Problem

Skills were learned once and trusted forever.

The reinforcement machinery already existed on `main` — confidence rose on
success, fell twice as fast on failure, and anything under 0.4 was "flagged".
Three things were missing:

1. **The flag was computed and then ignored.** `task_tracker begin` told the
   model *"⚡ SKILL MATCH FOUND — YOU MUST USE IT … do NOT proceed with a manual
   approach"* regardless of confidence. A skill that had failed 5 of 7 times got
   the same mandatory language as a perfect one. `isSkillFlagged` was only read
   by `status`/`review` — commands a human had to go and type.
2. **Failure lessons were write-only.** Each failure appended the model's own
   one-line excuse to `negative_examples`. Nothing ever read that list to change
   anything, so `## Steps` stayed wrong forever and the same mistake repeated.
3. **A broken skill stayed broken.** No repair path existed: hand-edit it,
   delete it, or keep following it — and per (1), the agent was being told to
   keep following it.

Net effect was a ratchet: a bad skill is mandated → fails → confidence drops →
still mandated → fails again, each failure writing a shallow lesson that nothing
consumes.

## What's in this PR

**1. Skills learn from failure** (`a5f1ba6`)
On any non-success outcome, `task_tracker end` replaces the model's one-line
excuse with a grounded root-cause + next-strategy reflection derived from the
steps actually recorded. Fails soft — a reflection error keeps the raw reason.

**2. Skills can repair themselves** (`a5f1ba6`)
New `skill_learner repair` reads a flagged skill's accumulated lessons and
rewrites its steps to avoid each one. Capped at 3 repairs per skill, after which
it requires `update`/`delete`. Confidence resets to 0.6 — enough to clear the
flag, not enough to be treated as proven.

**3. A human decides when it matters** (`396f3f3`)
- Flagged match → prompt **proceed / repair / skip**, with the skill's
  confidence, success/failure ratio, and recent failures shown.
- Proposed repair → **diff** of current vs proposed steps, then
  **accept / edit / cancel**. `edit` opens `$VISUAL`/`$EDITOR` and re-shows the
  diff of what you saved. Nothing is written or committed before accept.
- The user's answer reaches the model as an explicit `USER DECISION:`
  instruction, so it can't quietly do the opposite.
- Headless/SDK: one `autoRepair` boolean instead of prompts.

**4. Closed a hole this exposed** (`fix` commit)
`crystallize` could overwrite an existing skill, resetting confidence to 1.0 and
erasing its failure history with no approval. Found when a live voice session
did exactly that, unprompted, after being refused a repair.

## Behavior changes

- **Unattended repair now requires opt-in.** `autoRepair` defaults off, so voice
  sessions and scheduled jobs no longer auto-repair; they get a "flagged,
  repair is disabled" report instead. Rationale: an agent rewriting instructions
  it will then follow shouldn't happen unobserved. Flip the default in
  `sdk-types.ts` if we'd rather preserve the old behavior.
- **`crystallize` refuses an existing name** and points at `update`.

### DIFF
--- a/src/tools/skill-learner.ts
+++ b/src/tools/skill-learner.ts
@@ -1,4 +1,5 @@
 import { readFile, writeFile, mkdir, readdir, rm } from "fs/promises";
+import { existsSync } from "fs";
 import { join } from "path";
 import { execSync } from "child_process";
 import { type Static } from "@sinclair/typebox";
@@ -240,6 +241,25 @@ export function createSkillLearnerTool(
 					// Validate kebab-case
 					if (!/^[a-z0-9]+(-[a-z0-9]+)*$/.test(params.skill_name)) {
 						throw new Error("skill_name must be kebab-case (e.g., deploy-staging)");
+					}
+
+					// Crystallize writes a fresh SKILL.md with confidence 1.0 and empty
+					// stats. Doing that over an existing skill would erase its whole track
+					// record — including the failures that flagged it — and swap its
+					// instructions for one task's step log, with none of the approval a
+					// "repair" needs. Refuse and point at the non-destructive action.
+					if (existsSync(join(agentDir, "skills", params.skill_name, "SKILL.md"))) {
+						const existing = await loadSkillStats(join(agentDir, "skills", params.skill_name));
+						return {
+							content: [{
+								type: "text",
+								text: `Skill "${params.skill_name}" already exists (confidence ${existing.confidence}, ` +
+									`${existing.success_count} success / ${existing.failure_count} failure). Crystallize does NOT overwrite — ` +
+									`that would erase its confidence, usage history, and recorded failures.\n\n` +
+									`Use action "update" to change its steps, or crystallize under a different skill_name.`,
+							}],
+							details: { skill_name: params.skill_name, created: false, reason: "already_exists" },
+						};
 					}
 
 					const store = await loadTasks(gitagentDir);
--- a/src/tools/skill-learner.ts
+++ b/src/tools/skill-learner.ts
@@ -3,10 +3,20 @@ import { join } from "path";
 import { execSync } from "child_process";
 import { type Static } from "@sinclair/typebox";
 import type { AgentTool } from "@mariozechner/pi-agent-core";
+import type { Model } from "@mariozechner/pi-ai";
+import type { GCAssistantMessage } from "../sdk-types.js";
 import { skillLearnerSchema } from "./shared.js";
 import { loadSkillStats, isSkillFlagged } from "../learning/reinforcement.js";
+import { repairSkillSteps } from "../learning/skill-repair.js";
 import type { TaskRecord } from "./task-tracker.js";
 import yaml from "js-yaml";
+
+// Caps how many times a single skill can be auto-repaired before it must
+// go to a human (via "update" or "delete") instead.
+const MAX_REPAIRS = 3;
+// Clears the <0.4 flag immediately but stays below what a genuinely-proven
+// skill earns — a repaired skill has to re-earn trust, not start clean.
+const REPAIR_RESET_CONFIDENCE = 0.6;
 
 // ── Helpers ─────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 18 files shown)


src/tools/skill-learner.ts
```diff
@@ -342,11 +424,162 @@ export function createSkillLearnerTool(agentDir: string, gitagentDir: string): A
...
+						content: [{
+							type: "text",
+							text: `Flagged skills (confidence < 0.4):\n${lines.join("\n")}\n\nCall skill_learner action "repair" with a skill_name to let the agent rewrite its steps using these lessons (up to ${MAX_REPAIRS} times per skill). If a skill has already been repaired ${MAX_REPAIRS} times, use "update" or "delete" instead.`,
+						}],
 						details: { flagged },
 					};
 				}
 
+				case "repair": {
+					if (!params.skill_name) throw new Error("skill_name is required for repair action");
```
shreyas-lyzr, line 447: repair, update, and delete accept any string here. crystallize validates kebab-case on line 242 but these three actions do not. The file-existence check limits blast radius, but adding the same `/^[a-z0-9]+(-[a-z0-9]+)*$/` guard here and in update/delete would close the path-traversal window for prompt-injection scenarios.

shreyas-lyzr (approved):
This is a well-structured PR that addresses three real gaps in the skill reinforcement loop. The architecture is sound, the approval gate is correctly placed, and the test suite is thorough. A few things worth addressing before or shortly after merge:

**Correctness — repair/update/delete accept un-validated skill_name**

crystallize validates skill_name with `/^[a-z0-9]+(-[a-z0-9]+)*$/` (lines 242-243 of skill-learner.ts), but the repair, update, and delete actions skip that check. In practice, all extant skill names come from crystallize so they are already kebab-case — but an externally-crafted call (prompt injection, direct SDK use) could pass `skill_name = "../../../some/path"`. The file-existence check (readFile → "Skill not found") limits the damage to files that already exist, but defense-in-depth suggests adding the same regex guard to the other three actions. Not blocking given the internal nature of this tool, but worth a follow-up.

**Minor correctness — repair reads the SKILL.md twice**

The repair case reads the file once directly (line 440: `readFile(skillFile)`) to parse frontmatter and body, then calls `loadSkillStats(skillDir)` (line 453) which reads it a second time for `stats.negative_examples`. These two reads are sequential in a single-threaded async context, so there is no practical race, but a single parse call would be cleaner and remove the theoretical staleness window entirely.

**Subtle edge case — empty editor result is silently accepted**

If the user chooses `edit` and clears the entire file (saves an empty buffer), `edited.trim() === initial.trim()` is false, so the code sets `finalSteps = ""` and `userEdited = true`. The repair then writes a SKILL.md with an empty Steps section, which would silently break the skill. A one-line guard (`if (edited !== null && edited.trim().length > 0)`) would prevent this.

**Suggestion — the reflected failure reason might be verbose as a negative_example**

The reflection prompt caps output at 500 chars (2-4 sentences). That text is then stored verbatim as a `negative_example` entry, which in turn gets fed as input to the repair prompt (up to 10 examples × 500 chars = 5000 chars of injected context). The repair prompt already has its own cap at 3000 chars output — this won't cause errors — but a terse lesson (one sentence) would be more useful than a paragraph at retrieval time. Consider trimming the lesson before appending to `negative_examples`, or splitting the reflection output into a "lesson" field separate from the verbose analysis.

**Security pass**

No new dependencies added in this PR. No hardcoded secrets or keys found in any of the new files. No injection/authz gaps reachable from untrusted input (skill_name values are not user-supplied in normal flows). The `autoRepair` opt-in default-off is the right call — an agent rewriting instructions it will then follow unobserved is the correct thing to prevent by default.

**Tests**

The hermetic suite in learning.test.ts is solid: reinforcement math, the full task lifecycle, all guard rails. The Proxy-based exploding model is a good technique for proving the fail-soft paths were actually reached (the `touched` flag closes the gap). The live test file is gated correctly behind `GITAGENT_LIVE_TESTS`. No coverage gaps on the new paths that can be exercised without a real model.

src/tools/skill-learner.ts
Nivesh353: Thanks — all three actionable items are in, plus tests.

1. **skill_name validation** — extracted `assertValidSkillName` and applied it to
   `repair`, `update` and `delete`. I'd rate this above defense-in-depth: `delete`
   has no existence gate before `rm(dir, { recursive: true })`, so traversal
   reached outside the skills tree. Verified against the pre-fix build:

PR #105, opened 2026-08-14
feat(telemetry): one session id and one trace per turn (RPT-057)

Nivesh353:
What — Every model request now carries X-Session-Id, and every request within a turn shares a traceparent.

Why — RPT-057: one gitagent run produced several unrelated traces in Studio with no way to correlate them or total consumption.

How — loadAgent() takes an optional session id (CLI --session-id, SDK sessionId) and puts it on the model's headers, cloned so the shared pi-ai registry object isn't mutated. startTurnTrace() sets a fresh traceparent before each turn.

Verified — Against dev: two traces sharing one session id, then a single trace of 12 spans once the server-side change landed

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -53,7 +53,6 @@ interface ParsedArgs {
 	repo?: string;
 	pat?: string;
 	session?: string;
-	sessionId?: string;
 	voice?: string;
 }
 
@@ -69,7 +68,6 @@ function parseArgs(argv: string[]): ParsedArgs {
 	let repo: string | undefined;
 	let pat: string | undefined;
 	let session: string | undefined;
-	let sessionId: string | undefined;
 	let voice: string | undefined;
 
 	for (let i = 0; i < args.length; i++) {
@@ -110,11 +108,6 @@ function parseArgs(argv: string[]): ParsedArgs {
 			case "--session":
 				session = args[++i];
 				break;
-			// Distinct from --session (a git branch for repo/sandbox mode): this is
-			// the id carried on model requests so a gateway can group the run.
-			case "--session-id":
-				sessionId = args[++i];
-				break;
 			case "--voice":
 			case "-v":
 				// Accept optional backend name: --voice, --voice openai, --voice gemini
@@ -132,7 +125,7 @@ function parseArgs(argv: string[]): ParsedArgs {
 		}
 	}
 
-	return { model, dir, prompt, env, sandbox, sandboxRepo, sandboxToken, repo, pat, session, sessionId, voice };
+	return { model, dir, prompt, env, sandbox, sandboxRepo, sandboxToken, repo, pat, session, voice };
 }
 
 function handleEvent(
@@ -334,7 +327,7 @@ async function main(): Promise<void> {
 		return;
 	}
 
-	const { model, dir: rawDir, prompt, env, sandbox: useSandbox, sandboxRepo, sandboxToken, repo, pat, session: sessionBranch, sessionId: sessionIdFlag, voice } = parseArgs(process.argv);
+	const { model, dir: rawDir, prompt, env, sandbox: useSandbox, sandboxRepo, sandboxToken, repo, pat, session: sessionBranch, voice } = parseArgs(process.argv);
 
 	// If --repo is given, derive a default dir from the repo URL (skip interactive prompt)
 	let dir = rawDir;
@@ -478,7 +471,7 @@ async function main(): Promise<void> {
 
 	let loaded;
 	try {
-		loaded = await loadAgent(dir, model, env, sessionIdFlag);
+		loaded = await loadAgent(dir, model, env);
 	} catch (err: any) {
 		console.error(red(`Error: ${err.message}`));
 		process.exit(1);
--- a/src/index.ts
+++ b/src/index.ts
@@ -312,6 +312,11 @@ async function ensureRepo(dir: string, model?: string): Promise<string> {
 	return absDir;
 }
 
+// The REPL outlives main(): main() resolves once the prompt loop is wired up, so
+// telemetry must be flushed by whichever exit path the user actually takes, not
+// when main()'s promise settles.
+let _replActive = false;
+
 async function main(): Promise<void> {
 	// Handle plugin subcommand: gitagent plugin <install|list|remove|...>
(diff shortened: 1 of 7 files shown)


src/telemetry.ts
```diff
@@ -184,6 +185,33 @@ export function isTelemetryEnabled(): boolean {
...
 
+// ── Turn-scoped trace propagation ──────────────────────────────────────
+
+/**
+ * Start a new W3C trace for one user turn.
+ *
+ * A single user message costs several HTTP calls to the model gateway — one
+ * that comes back with a tool call, another with the answer, and so on. Each
+ * call is a separate request, so a gateway that traces per request records one
+ * trace per call and the turn arrives split across several of them. Sending the
```
shreyas-lyzr, line 196: Guard is inverted. `_initialized` is `true` when OTel is active, so this returns early when telemetry is on and writes a header when it is off — the opposite of what the docstring describes. Fix:

```suggestion
	if (!_initialized || !model) return;
```

shreyas-lyzr (commented):
Five focused items across correctness, logic, and security — two worth addressing before ship.

**1. Guard in startTurnTrace is inverted (logic bug)**

The guard is `if (_initialized || !model) return;` — this makes the function a no-op when OTel is active (`_initialized === true`) and writes a manual `traceparent` when telemetry is *disabled*. That is the opposite of the documented intent. The docstring says: when OTel is initialised, `UndiciInstrumentation` already injects `traceparent` from the active span automatically and a manually-written header would fight it — so skip when enabled, act when disabled. The correct guard is `if (!_initialized || !model) return;`.

**2. startTurnTrace mutates the model object in place across turns**

`m.headers = { ...(m.headers ?? {}), traceparent }` writes directly to `loaded.model`. loader.ts explicitly clones the registry object to prevent exactly this kind of cross-run contamination. In the REPL each turn overwrites the previous traceparent so it works serially, but in a concurrent SDK scenario (two `query()` calls in the same process) both writers race on the same object. Consider returning a headers delta and letting the caller compose it, or documenting the single-threaded assumption clearly.

**3. _replActive declaration sits below the call that references it**

`let _replActive = false` is declared after `main()` is called at module level. With `let` there is no hoisting, so the declaration is reached only because `main()` is async and `.finally()` runs after the event loop tick. It works in practice but is fragile to read. Moving the declaration above `main()` costs one line and removes the confusion.

**4. Telemetry flush on single-shot error path — appears safe, but non-obvious**

When `agent.prompt()` throws in single-shot mode, the error propagates through `main().catch()` which calls `process.exit(1)` — but the `.finally(() => { if (!_replActive) shutdownTelemetry() })` on the chain fires first, so telemetry is flushed. This is correct; a brief comment noting it is intentional would save the next reader from worrying about it.

**5. Security pass: clean**

No new runtime dependencies. The only new import is `randomBytes` from Node's built-in `crypto`. `X-Session-Id` carries a UUID; `traceparent` is a W3C correlation header — neither is a credential. No injection, SSRF, IDOR, or authz surface touched. No secrets in the diff.

src/telemetry.ts
```diff
@@ -184,6 +185,33 @@ export function isTelemetryEnabled(): boolean {
...
+ * same `traceparent` on every call of the turn lets the gateway stitch them
+ * into one trace.
+ *
```
shreyas-lyzr, line 199: Mutates `loaded.model.headers` in place. The loader clones the registry object for exactly this reason (see the comment above the spread in loader.ts). Works in the single-threaded REPL because each turn overwrites the previous `traceparent`, but two concurrent `query()` calls in the same process race here. Consider returning the new headers map rather than writing through to the shared object.

src/index.ts
```diff
@@ -843,20 +854,31 @@ async function main(): Promise<void> {
...
 		}
 	});
 
+	_replActive = true;
 	ask();
 }
 
+// The REPL outlives main(): main() resolves once the prompt loop is wired up,
+// so telemetry must be flushed by whichever exit path the user actually takes,
+// not when main()'s promise settles.
```
shreyas-lyzr: `let _replActive = false` is declared after `main()` is invoked at module level. It happens to work because `main()` is async and `.finally()` runs later, but declaring it above the `main()` call would make the dependency order explicit.

src/telemetry.ts
Nivesh353: The write-through is deliberate. Agent is constructed with `model: loaded.model` and
pi-ai reads `headers` at request time, so a copy made here would never be sent —
returning a headers map would leave the caller writing to the same object anyway.

It's safe because the model is already this run's own: loadAgent clones it off the
shared registry, and every query() loads its own, so concurrent runs never share one.
Turns within a run are sequential, so this is the only writer per object.

Added all of that to the docstring, since you're right that it wasn't written down.

src/index.ts
Nivesh353: Moved it above main(). Small correction though: it was already declared before main()
was invoked (line 870 vs 877), so there was no TDZ or async-timing dependency — but it
sat below the function body that reads it, which reads confusingly. Now it's above.

src/telemetry.ts
Nivesh353: Not changing this one — I think the suggestion inverts it the wrong way.

Telemetry is off by default, so with `if (!_initialized)` we'd never write the
traceparent on the normal path and every call would get its own trace, which is the
bug this PR fixes. And when OTel *is* on, it would write a header competing with
UndiciInstrumentation — the thing the docstring warns about.

The current guard matches the intent as you described it: skip when enabled, act when
disabled. Verified on a real run with telemetry off — one trace, 12 spans, all sharing
the trace id, which only happens because this code runs.

shreyas-lyzr (approved):
All three findings from the previous round are addressed.

The logic in startTurnTrace (if (_initialized || !model) return) was not a bug — the new JSDoc makes the intent unambiguous: when OTel is active, undici already injects traceparent from the active span, so this function deliberately yields to it. The guard is correct.

The write-through to model.headers is now clearly documented as safe: loadAgent clones the model off the shared registry, every query() load gets its own copy, concurrent runs never share an object, and turns within a run are sequential. The mutation is intentional.

_replActive is now hoisted above main(), with a comment explaining why the REPL outlives main()'s promise.

Security pass: no changes to deps, no secrets, no injection/authz surface touched. Clean.

PR #109, opened 2026-09-09
feat(version): tag, diff and roll back an agent's config

Nivesh353:
## Summary

An agent is already a git repo, so a known-good configuration is just a tag — but restoring one meant hand-running git plumbing and knowing which paths are safe to touch. This adds an ergonomic CLI over those primitives. No new storage: versions are annotated git tags in the agent's own repo.

```bash
gitagent version save v1.0 -m "baseline before model swap"
gitagent version list
gitagent version show v1.0
gitagent version diff v1.0            # vs working tree
gitagent version diff v1.0 v1.1       # version vs version
gitagent version rollback v1.0 --dry-run
gitagent version rollback v1.0
```

| Subcommand | Description |
|---|---|
| `save [<name>]` | Tag the current config. Name defaults to `v` + `version` from `agent.yaml` |
| `list` | List saved versions (`--json` for scripting) |
| `show <name>` | Version details, file inventory, and drift since |
| `diff <a> [<b>]` | Config diff between versions, or against the working tree |
| `rollback <name>` | Restore config from a version as a new commit |

## Three design decisions worth reviewing

**1. Rollback is config-only — `memory/` and `skills/` are never touched.**

`src/tools/memory.ts` commits on *every* memory save and `skill_learner` rewrites `skills/` at runtime, so those commits interleave with config commits throughout history. A whole-tree restore would silently destroy everything the agent learned after the tag was cut.

Versioned paths: `agent.yaml`, `SOUL.md`, `RULES.md`, `DUTIES.md`, `AGENTS.md`, `config/`, `tools/`, `hooks/`, `knowledge/`, `examples/`, `compliance/`, `agents/`, `workflows/`, `schedules/`, `plugins/`.

**2. Rollback is forward-only.**

It writes a *new* commit rather than rewriting history, so the rollback is visible in `git log` and revertible like anything else. No `reset --hard`, no branch switching — safe with a remote or other clones.

**3. Restore is diff-planned, not a blanket checkout.**

A plain `git checkout <tag> -- tools/` leaves files added *after* the tag in place, producing a merge of old and new config rather than a rollback. Instead the plan comes from `git diff --name-status --no-renames -z <tag> HEAD -- <pathspec>`, then remove-before-restore (that ordering also handles file↔directory flips). Because every path is computed by git under the config pathspec, it is structurally impossible for `memory/` to enter the plan.

## Also adds `src/git.ts`

Every call is `execFileSync` with an argv array, so the quoting/injection bug class is structurally impossible. It also handles `--no-pager` (otherwise `diff` opens a pager and appears to hang), a 64 MiB buffer, closed stdin, and per-invocation identity + `gpgsign=false` so commits work in a repo with no `user.email` without ever writing to the user's git config.

Scope note: only `isGitRepo` is migrated to it. The other 13 ad-hoc `execSync` git sites — including the unescaped commit message at `src/session.ts:138` — are left alone deliberately; they are on untested hot paths, and porting them belongs in a follow-up whose diff reads as a refactor rather than being buried in a feature.

## Test plan

- [x] `npm run build` clean; `npm test` 94 passing (29 new)
- [x] Establishes the temp-git-repo fixture the suite lacked, hermetic via `GIT_CONFIG_GLOBAL=/dev/null` so CI and dev machines agree
- [x] Shell-injection regression test for the bug class live in `src/tools/memory.ts`
- [x] Verified end-to-end on a real 276-commit agent repo: config reverted, file added after the tag removed, **memory written after the tag survived**, history forward-only with the tag intact
- [x] Guard rails verified: dirty config, duplicate tag, unknown version, no-op rollback (creates no empty commit), detached HEAD, staged index
- [x] `gitagent plugin` still works after the shared `--dir` extraction

## Docs

README gains a `## Versioning & Rollback` section; `Documentation.md` gains a full `### Version CLI` reference with edge-case behavior and known limitations.

## Out of scope

`--push` / remote tag sync, versioning `memory/` or `skills/`, per-agent tag namespacing (`agentcfg/<agent>/<version>`), and porting the remaining `execSync` sites. Tags stay local until pushed manually — `save` prints the exact command.

### DIFF
--- a/src/version-cli.ts
+++ b/src/version-cli.ts
@@ -1,0 +1,270 @@
+import { createInterface } from "readline";
+import { createRequire } from "module";
+import { gitStream } from "./git.js";
+import {
+	TAG_NS,
+	VersionError,
+	applyRollback,
+	assertRollbackReady,
+	buildDiffArgs,
+	hasConfigDiff,
+	isShallow,
+	listVersions,
+	planRollback,
+	resolveContext,
+	saveVersion,
+	showVersion,
+} from "./version.js";
+import type { RollbackPlan, VersionContext } from "./version.js";
+
+const require = createRequire(import.meta.url);
+const { version: GITAGENT_VERSION } = require("../package.json");
+
+const dim = (s: string) => `\x1b[2m${s}\x1b[0m`;
+const bold = (s: string) => `\x1b[1m${s}\x1b[0m`;
+const red = (s: string) => `\x1b[31m${s}\x1b[0m`;
+const green = (s: string) => `\x1b[32m${s}\x1b[0m`;
+
+function flagValue(args: string[], ...names: string[]): string | undefined {
+	for (const name of names) {
+		const i = args.indexOf(name);
+		if (i !== -1 && args[i + 1]) return args[i + 1];
+	}
+	return undefined;
+}
+
+/** Positional args: everything that isn't a flag or a flag's value. */
+function positionals(args: string[], valued: string[]): string[] {
+	const out: string[] = [];
+	for (let i = 0; i < args.length; i++) {
+		if (valued.includes(args[i])) {
+			i++;
+			continue;
+		}
+		if (args[i].startsWith("-")) continue;
+		out.push(args[i]);
+	}
+	return out;
+}
+
+function confirm(question: string): Promise<boolean> {
+	return new Promise((resolve) => {
+		const rl = createInterface({ input: process.stdin, output: process.stdout });
+		rl.question(`${question} [y/N] `, (answer) => {
+			rl.close();
+			resolve(/^y(es)?$/i.test(answer.trim()));
+		});
+	});
+}
+
+function printPaths(paths: string[], limit = 20): void {
+	for (const p of paths.slice(0, limit)) console.log(`    ${p}`);
+	if (paths.length > limit) console.log(dim(`    …and ${paths.length - limit} more`));
+}
+
+function warnIfShallow(ctx: VersionContext): void {
+	if (isShallow(ctx)) {
+		console.log(dim("Note: shallow clone — some tags may be missing. git fetch --unshallow --tags"));
+	}
+}
+
+// ── Subcommands ────────────────────────────────────────────────────────
+
+async function handleSave(agentDir: string, args: string[]): Promise<void> {
+	const ctx = resolveContext(agentDir);
+	const name = positionals(args, ["-m", "--message"])[0];
+	const result = saveVersion(ctx, {
+		name,
+		message: flagValue(args, "-m", "--message"),
+		commit: args.includes("--commit"),
+		force: args.includes("--force"),
+	});
+
+	if (result.committed.length > 0) {
+		console.log(dim(`Committed ${result.committed.length} config file(s) first.`));
+	}
+	console.log(`${green("Saved version")} ${bold(result.name)} ${dim(`(${result.commit})`)}`);
+	console.log(dim(`${result.fileCount} config files tracked · memory/ and skills/ not included`));
+	console.log(dim(`Local only — push with: git push origin ${result.ref}`));
+}
+
+function handleList(agentDir: string, args: string[]): void {
+	const ctx = resolveContext(agentDir);
+	const versions = listVersions(ctx);
+
+	if (args.includes("--json")) {
+		console.log(JSON.stringify(versions, null, 2));
+		return;
+	}
+	if (versions.length === 0) {
+		console.log(dim("No versions saved yet. Create one with: gitagent version save v1.0"));
+		return;
+	}
+	const width = Math.max(...versions.map((v) => v.name.length));
+	for (const v of versions) {
+		const meta = dim(`${v.date.slice(0, 10)}  ${v.commit}`);
+		console.log(`  ${bold(v.name.padEnd(width))}  ${meta}  ${v.subject}`);
+	}
+	warnIfShallow(ctx);
+}
+
+function handleShow(agentDir: string, args: string[]): void {
+	const ctx = resolveContext(agentDir);
+	const name = positionals(args, [])[0];
+	if (!name) {
+		console.error(red("Usage: gitagent version show <name> [--files]"));
+		process.exit(1);
+	}
+	const info = showVersion(ctx, name);
+
+	console.log(`${bold(info.name)} ${dim(`(${info.commit})`)}`);
+	if (info.message) console.log(info.message);
+	console.log(dim(`${info.author} · ${info.date}`));
+	console.log();
+	console.log(dim(`${info.files.length} config file(s) at this version:`));
+	printPaths(info.files, args.includes("--files") ? info.files.length : 20);
+	console.log();
+	if (info.driftStat) {
+		console.log(dim("Changes since this version:"));
+		console.log(info.driftStat);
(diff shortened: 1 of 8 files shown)


src/version.ts
```diff
@@ -0,0 +1,461 @@
...
+	const body =
+		`Restored ${plan.restore.length} file(s), removed ${plan.remove.length} file(s) ` +
+		`from ${TAG_NS}/${plan.name} (${plan.commit}).`;
+	try {
+		// git rm and git checkout -- <paths> both stage their work, and the index was
+		// verified clean beforehand, so a bare commit captures exactly this plan.
+		gitWrite(ctx.root, ["commit", "-m", subject, "-m", body]);
+	} catch (err: any) {
+		throw new VersionError(
+			"COMMIT_FAILED",
```
shreyas-lyzr, line 438: The COMMIT_FAILED error message assumes the index is fully staged ('Changes are staged but not committed. Run git commit to finish'), but that is only true when the commit itself is the thing that failed.

If a restore batch fails mid-way — e.g., git checkout aborts on a corrupt object after the rm batch has already run — then only part of the plan is staged. Committing at that point produces a commit with some files deleted but not all files restored. The user would need to know what git reset --hard HEAD actually recovers, but the message presents commit as a valid option on equal footing.

Suggestion: distinguish the two cases. Wrap each git rm / git checkout call with its own try/catch; if one of them throws, the message should be specific: 'Files were partially modified — nothing is committed. Run git reset --hard HEAD to recover.' The current messaging only applies cleanly when the bare git commit call is the only thing that failed, which happens to be the most common case but is not the only one.

src/version.ts
```diff
@@ -0,0 +1,461 @@
...
+		else restore.push(path);
+	}
+	restore.sort();
+	remove.sort();
+	return { name, commit: git(ctx.root, ["rev-parse", "--short", `${ref}^{commit}`]), restore, remove };
+}
+
+export interface RollbackOptions {
+	message?: string;
+	force?: boolean;
```
shreyas-lyzr: RollbackOptions.force is declared here but never read anywhere inside applyRollback. The force semantics for rollback (skip dirty-config check, allow detached HEAD) live entirely in assertRollbackReady, which is called by the CLI handler before applyRollback is ever invoked.

This means a caller using the applyRollback API directly would see a force field that has no effect, which is confusing and could lead to misuse. Either remove the field from the interface, or move the assertRollbackReady call inside applyRollback so force actually does something there.

test/version.test.ts
```diff
@@ -0,0 +1,363 @@
...
+
+		// Forward-only: one new commit, nothing rewritten, still on main.
+		assert.equal(raw(root, "rev-list", "--count", `${before}..HEAD`), "1");
+		assert.doesNotThrow(() => raw(root, "merge-base", "--is-ancestor", before, "HEAD"));
+		assert.equal(raw(root, "rev-parse", "--abbrev-ref", "HEAD"), "main");
+
+		await cleanup(root);
+	});
+
+	it("is a no-op when config already matches, with no empty commit", async () => {
```
shreyas-lyzr, line 241: There is a test for INDEX_DIRTY during saveVersion (line 155) but no equivalent for the rollback path. assertRollbackReady also calls assertIndexClean, but that code path has no test coverage. Adding a case that stages a file and then calls assertRollbackReady would close the gap — especially since a staged index during rollback leaves the repo in a confusing state.

src/version-cli.ts
```diff
@@ -0,0 +1,270 @@
...
+	const extra: string[] = [];
+	if (args.includes("--stat")) extra.push("--stat");
+	if (args.includes("--name-only")) extra.push("--name-only");
+
+	if (!hasConfigDiff(ctx, a, b)) {
+		console.log(dim("No config differences."));
+		return;
+	}
+	console.log(dim(`${a} → ${b ?? "working tree"}`));
+	gitStream(ctx.root, buildDiffArgs(ctx, a, b, extra));
```
shreyas-lyzr, line 155: gitStream returns the git exit status but the return value is ignored here. git diff exits 1 when differences exist, so a non-zero exit is expected and harmless (the hasConfigDiff guard above already confirmed there is a diff). A comment noting this would make it clear the ignored return is intentional rather than an oversight, since ignoring an exit code is normally a yellow flag.

shreyas-lyzr (commented):
Solid feature with well-thought-out design decisions. The injection-safe git layer, the config-only scope, and the forward-only rollback are all correct. Four findings below, one of which is a real correctness bug in error handling.

**Security pass:** no new dependencies introduced, all git invocations go through execFileSync with argv arrays (no shell), version names are validated against a strict allowlist regex before any git call touches them. No CVEs, no secrets, no injection surfaces in the new code. Clean.

The unescaped execSync at src/session.ts:138 (acknowledged out of scope) remains the highest-risk site in the repo — worth a dedicated follow-up.

src/version.ts
Nivesh353: Good catch, and it was actually worse than described: a failure in the rm/checkout phase threw a raw GitError that escaped the CLI's VersionError handler, so the user got a stack trace and no guidance at all while the repo sat half-mutated.
Split into two codes — RESTORE_FAILED ("partially modified, nothing committed, reset --hard to recover") and COMMIT_FAILED ("complete rollback is staged, git commit to finish") — and both now warn that reset --hard also discards uncommitted memory/, which the original message glossed over. Verified by forcing a checkout failure on a read-only directory.

src/version.ts
Nivesh353: Correct, dead field — force is entirely consumed by assertRollbackReady before applyRollback is ever called. Removed it from the interface and the CLI call site.
Went with removal rather than moving assertRollbackReady inside applyRollback, since the CLI prints the plan and prompts for confirmation between those two steps; left a one-line comment at the call site so it doesn't get re-added.

test/version.test.ts
Nivesh353: Added — stages a memory/ file and asserts assertRollbackReady throws INDEX_DIRTY.
It also pins down something worth locking in: --force covers dirty config and detached HEAD but deliberately does not override a dirty index, since that's the guard stopping staged memory/ from riding along in the rollback commit.

src/version-cli.ts
Nivesh353: Fair — an ignored exit code should never be silent. Added a comment noting git diff exits 1 when differences exist, and that hasConfigDiff above already established there are some.
No behavior change.

shreyas-lyzr (approved):
All four findings addressed — the follow-up commit is clean.

**RESTORE_FAILED vs COMMIT_FAILED split** — exactly right. The try/catch wrapping the rm+checkout phase now catches partial-restore failures and throws a distinct `RESTORE_FAILED` error with accurate guidance. The commit path stays as `COMMIT_FAILED`. Both messages now note that `reset --hard` also discards uncommitted `memory/`, which the old message glossed over. Better than the minimal fix I suggested.

**`RollbackOptions.force` removed** — correct call. Removal (not dead-field documentation) is the right answer; the field was never semantically part of `applyRollback`'s contract. The CLI call site is updated accordingly.

**INDEX_DIRTY rollback test added** — the new test covers both `force=false` and `force=true`, and explicitly locks in that `--force` does *not* bypass a dirty index. That's an important behavioral contract to pin in tests.

**`gitStream` exit status comment** — the comment is accurate and saves the next reader from the same question.

95 tests passing. No new concerns. Approving.

PR #112, opened 2026-09-17
fix(telemetry): preserve tool-call trace correlation

### LINKED ISSUE #26: GitClaw traces not showing skill/tool calls in Lyzr Agent Studio (only high-level logs visible)
**Problem**
In Studio, GitClaw traces are not showing detailed execution steps such as:
Skill calls
Tool calls
Intermediate agent actions

The trace timeline only displays high-level steps like:
Session
Agent orchestration
Generate AI response

Even in Detailed Logs, the breakdown of actual tool/skill execution is missing. This makes it difficult to understand what the agent is actually doing internally.

**Expected Behavior**
Trace view in Studio should include:

- Skill-level execution steps
- Tool calls with inputs/outputs
- Intermediate reasoning/actions (if applicable)
- Clear step-by-step breakdown of agent flow

swsarancodes:
## Summary

Fixes #26.

- Preserve pi-agent-core’s runtime tool-call ID instead of generating an unrelated random ID.
- Add standard `gen_ai.operation.name`, `gen_ai.tool.name`, `gen_ai.tool.type`, and `gen_ai.tool.call.id` attributes so OpenTelemetry backends can classify and correlate tool executions.
- Preserve existing `tool.*` attributes for dashboard compatibility.
- Keep tool arguments and results out of spans to retain the existing privacy boundary.

## Before / after

| Before | After |
| --- | --- |
| The wrapper read the call ID from the second argument even though pi-agent-core passes it first, so every span received a random ID. | Each span carries the actual engine tool-call ID in both `tool.call_id` and `gen_ai.tool.call.id`. |
| Tool spans only exposed project-specific attributes. | Tool spans include standard GenAI operation and tool attributes that trace UIs recognize. |

## Validation

- `npm run build`
- `npm test` (76 passing)


swsarancodes:
Closing this telemetry-only change to replace it with a user-visible localhost fix and real before/after verification.

